PR #1, opened 2018-09-24
http server and manager modification

JackTemaki:
The _sis_vis_name attribute of a "Job" class can now be used to show custom names in the manager view.

The printing of job-errors inside the manager view can be turned off.

When encapsulating jobs with a sub_block, e.g.

`block = tk.sub_block("some_block")`
`with block:`
`some job-calls....`

This blocks graph can be rendered in the "Visualize Dependency Graph" section of the http server view.

Additionally, the http job view now displays job outputs.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -172,6 +172,8 @@ WAIT_PERIOD_MTIME_OF_INPUTS = 60
 
 #: set true to automatically clean jobs in error state and retry
 CLEAR_ERROR = False
+
+PRINT_ERROR = True
 #: Print detailed log of that many jobs in error state
 PRINT_ERROR_TASKS = 1
 #: Print that many last lines of error state log file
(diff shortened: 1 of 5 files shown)


sisyphus/global_settings.py
```diff
@@ -172,6 +172,8 @@ def file_caching(path):
 
 #: set true to automatically clean jobs in error state and retry
 CLEAR_ERROR = False
+
```
critias: Add a comment starting with '#:' before PRINT_ERROR, otherwise it will not show up in the sphinx documentation.

sisyphus/http_server.py
```diff
@@ -268,6 +269,7 @@ def visualize(block_id):
     url_prefix = '/vis/' + block_id
 
     if parent is None:
+        items += all_root_blocks[0].get_sub_blocks()
```
critias: Why are only sub_blocks of the first root_block added? Better would be
root_block_a
sub_block_a1
sub_block_a2
...
root_block_b
sub_block_b1
...

PR #3, opened 2018-10-18
added manager callback function to display dynamic job information

JackTemaki:
There was a request to add dynamic information into the manager view. You can override the function to provide additional dynamic information about the job, which can e.g. be the perplexity of a running model training job.

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -1051,6 +1051,14 @@ class Job(object, metaclass=JobSingleton):
         to be displayed on the web interface """
         pass
 
+    def manager_callback(self):
+        """ Is executed each time the job is displayed in the manager,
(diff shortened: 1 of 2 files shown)


JackTemaki:
Should I use the already available "info" function? I am not sure how it is used in the web interface already and I guess it can be used for multiline information, while this function should return single line. so maybe short_info?

critias:
Good point, info is used for exactly that purpose, it can currently handle arbitrary object and show them on the web interface. It's currently not used for anything but strings as far as I know. Take info and adjust the doc that it should return a string. Info is the more general name and this way the web interface will automatically show the information as well. If some additional detailed information is wanted for the web interface we introduce a new, more specific function.

PR #5, opened 2018-11-19
print job info only when job is running

JackTemaki:


sisyphus/manager.py
```diff
@@ -254,7 +254,7 @@ def print_state_overview(self, verbose=False):
                 if hasattr(job, "get_vis_name") and job.get_vis_name() is not None:
                     info_string += " [%s]" % job.get_vis_name()
 
-                if hasattr(job, "info"):
+                if hasattr(job, "info") and state in [gs.STATE_RUNNING]:
```
critias: Use state == gs.STATE_RUNNING

PR #10, opened 2018-12-10
too much parameters when determining deletions

JackTemaki:
the 'du' command has a parameter limit, which can be exceeded when checking too many folders.
2500 is an estimate, open to discussion.


critias:
Please add a warning if this workaround is used and a todo comment before it. The last summary given by 'du', the one everybody will directly see, will show not show the correct size of all files that would be deleted. It will only show the size of the last block.

PR #11, opened 2018-12-13
added shortcuts for sisyphus graph search, fixed time=0 issue

JackTemaki:


sisyphus/tools.py
```diff
@@ -60,7 +60,7 @@ def str_to_hours(t):
         assert(len(t) == 3)
         t = int(t[0]) * 3600 + int(t[1]) * 60 + int(t[2])
         t /= 3600
-    return t
+    return t if t > 0 else 1
```
critias: The better solution is to change line 57 to t = float(t)

JackTemaki:
I thought the int had some reason, then i will also change the rest accordingly

sisyphus/tools.py
```diff
@@ -54,14 +54,13 @@ def str_to_hours(t):
...
-        t = int(t)
+        t = float(t)
     except ValueError:
         t = t.split(':')
         assert(len(t) == 3)
         t = int(t[0]) * 3600 + int(t[1]) * 60 + int(t[2])
-        t /= 3600
-    return t if t > 0 else 1
-
+        t /= 3600.0
```
critias: This wasn't necessary, in python3 't /= 3600' already returned a float. On the other hand I guess it's more intuitive this way.

PR #12, opened 2019-01-08
 Show 'already used output' warning only if target changes 

JackTemaki:
The warning did never indicate if there was a real conflict, or only the same object registered twice. I think its nicer this way.


sisyphus/graph.py
```diff
@@ -230,16 +230,20 @@ def output(self):
 
     def add_target(self, target):
         """
-        :param OutputTarget target:
```
critias: add_target should handle all OutputTargets and not only path. It's currently also called by register_report and register_callback.

sisyphus/graph.py
```diff
@@ -230,16 +230,20 @@ def output(self):
...
         """
         self._targets.append(target)
 
         # check if output path is already used
         try:
-            path = target._output_path
-            if path in self.used_output_path:
-                logging.warning('Output path is used more than once: %s' % path)
-            self.used_output_path.add(path)
+            path = str(target._output_path)
```
critias: No need for str(..), this object should already contain a string

sisyphus/graph.py
```diff
@@ -230,16 +230,20 @@ def output(self):
...
+            if path in self.used_output_sis_path.keys() and self.used_output_sis_path[path] != target._sis_path:
```
critias: 'if path in self.used_output_sis_path.keys()' should only be  'if path in self.used_output_sis_path'

sisyphus/graph.py
```diff
@@ -188,7 +188,7 @@ def __init__(self):
         self._targets = []  # type: list[OutputTarget]
         self._active_targets = []  # type: list[OutputTarget]
         self._pool = None
-        self.used_output_path = set()
+        self.used_output_sis_path = dict()
```
critias: Keep the old name since it should not only handle sis_path

critias:
Feature was added in a different way

PR #13, opened 2019-01-11
Fixed issue with argument length during the directory size computation

ChristopherBrix:
By storing the list of directories in a temporary file, the number of directories no longer influences the command length.


critias:
Looks good, thanks

PR #16, opened 2019-01-31
added optional stacktrace to jobs, can be used in console

JackTemaki:
I thought this to be a very useful thing, it can be used to determine where a job was created.
Also, you can find all jobs belonging to a single call, e.g:

```
for job in j:
    first_st = job._sis_stacktrace[0][0]
    if first_st[0] == "/u/rossenbach/experiments/switchboard_test/config/E2ESystem.py" and first_st[1] == 229:
        print(job)
```

will list all jobs that are created in E2ESystem.py line 229:

```
225: j = CRNNSprintSearchJob(search_csp=self.csp[search_corpus],
226:                                feature_flow=self.feature_flows[feature_corpus][feature_flow],
227:                                crnn_config=crnn_config,
228:                                num_classes=self.functor_value(num_classes),
229:                                **kwargs)
```

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -254,7 +254,7 @@ VIS_ABSOLUTE_MERGE_THRESHOLD = 5
 SHOW_VIS_NAME_IN_MANAGER = True
 
 # Add stacktrace information with specified depth, 0 for deactivation
-JOB_ADD_STACKTRACE_WITH_DEPTH = 1
+JOB_ADD_STACKTRACE_WITH_DEPTH = 0
 
 # Internal functions
 def update_global_settings_from_text(text, filename):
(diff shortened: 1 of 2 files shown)


curufinwe:
I like the idea, but it probably should be disabled by default (stacktrace of depth 1 is not that useful anyway [depending on how you write your code]).

critias:
There is already an LOG_TRACEBACKS option to due something fairly similar. We should only have one of them. Your approach looks more flexible. I'm ok with keeping it, but in this case remove the old code.

And yes, disabling it by default is the right choice since adding a stack trace slows down the reading of the config file.

JackTemaki:
there was some weird git stuff going on when trying to merge...

JackTemaki:
ok, i removed the weird merges by a rebase. should be ready for pull now...

critias:
I don't think you wanted to add the .history files and sis.sh :)

JackTemaki:
jep, is fixed....

PR #17, opened 2019-02-04
Enable priorities using qsub job_share

vieting:
To allow assigning user-internal priorities to jobs, a `priority` option was added to the job requirements. It is passed to qsub's job_share option. As job_share is an unsigned integer with a default value of 0 and we want to allow decreasing a job's priority relative to other jobs, the default value in the job requirements is set to 50.

### DIFF
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -149,6 +149,10 @@ class SonOfGridEngine(EngineBase):
 
         out.append('-l')
         out.append('h_rt=%s' % task_time)
+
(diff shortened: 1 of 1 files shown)


critias:
I like the idea of having a priority keyword for each task, but this is currently a special case which is purely tailored towards SGE and directly exposes job_shares as priorities. I think a cleaner interface for priorities would be a range between -100 and 100 with 0 being the default.
If you only need it right now for the SGE you could already to it by setting qsub_args in the default_rqmt of the SGE and for the affected tasks.

PR #18, opened 2019-03-11
fixed issue with wrong return value of get_unfinished_jobs

JackTemaki:
A job can be finished without being runnable (cleanup deletions of inputs), leading to runnable jobs that are not required. This is solved by this fix.


critias:
Thanks

PR #19, opened 2019-03-12
Added flag to optionally ignore jobs in error state when the manager is started

ChristopherBrix:
In combination with the `-r` flag, this ensures that the manager starts whether there are jobs in the error state or not.


critias:
Thanks

PR #21, opened 2019-03-20
Added hashing support for namedtuples

ArneNx:
This adds hashing support for namedtuples. 
Previously namedtuples were treated as normal objects, which breaks down as soon as you have more complex objects, like e.g. a dict stored inside the namedtuple. 

An alternative to adding a separate method for checking the type would be to use `isinstance(tuple)`, but this might have some unwanted side-effects.

### DIFF
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -69,6 +69,23 @@ def get_object_state(obj):
         return args, state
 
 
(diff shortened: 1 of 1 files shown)


ArneNx:
Seems like I was wrong and this is not needed. 
The default behaviour (recursively going through the object state) works fine here.

PR #24, opened 2019-05-25
Avoid f-strings to ensure compatibility to python versions < 3.6

ChristopherBrix:
Commit ec7de9b1eb7d04dc38d48f0e8628ea716e50026c introduced f-strings, which are not supported for python versions older than 3.6. By using %-formatting instead, we don't lose support for older python versions.


critias:
I really like f-strings, but let's try to support Python 3.5 for a while longer.

PR #27, opened 2019-06-27
more verbose find_job function for better usability

JackTemaki:


critias:
I don't like that the return value changes. How about splitting this into two functions?
Function 1: find_jobs, it would basically be the current function, but always returns a list of jobs.
Function 2: find_job, returns always the last or first job of find_jobs output.

critias:
Actually, I don't think find_job would be needed in that case.
tk.find_jobs('...').pop() would do that already.

JackTemaki:
The two different return values are a problem, but I was annoyed that I always had to use something like j[0] (or pop) instead of the value returned. I am not sure what to do, if you do not want it in the repo, i will keep it locally and just push the more verbose print statements.

PR #30, opened 2019-08-08
added possible stderr capturing in tools.sh()

JackTemaki:


JackTemaki:
P.S: This is needed when calling ffmpeg jobs from sisyphus, as ffmpeg prints additional information into stderr, and not stdout. I use this to capture the new duration of the edited audio file.

sisyphus/tools.py
```diff
@@ -208,7 +209,8 @@ def sh(command,
 
     try:
         if capture_output:
-            return subprocess.check_output(command, shell=True, executable=executable).decode()
+            return subprocess.check_output(command, shell=True, executable=executable,
+                                           stderr=subprocess.STDOUT if include_stderr else subprocess.DEVNULL).decode()
```
critias: The current default behaviour is that not captured output still written to the logfile. Setting stderr=subprocess.DEVNULL will suppress it. Let's stay with stderr=None if it's include_stderr is false.
LGTM beside that.

JackTemaki:
I also added a quick fix for the `_sis_update_possible` function that breaks all setups including `update` calls

PR #35, opened 2020-01-16
ask for automatic creation of job when running worker

JackTemaki:
Sometimes when you run a job manually with the worker you forgot to actually create the job over the console first. As his happens more often, I though it might be nice to try to create the job if it does not exist when running the worker.

### DIFF
--- a/sisyphus/worker.py
+++ b/sisyphus/worker.py
@@ -189,8 +189,22 @@ def worker_helper(args):
             subprocess.check_call(call, stdout=logfile, stderr=logfile)
         return
 
-    with gzip.open(os.path.join(args.jobdir, gs.JOB_SAVE)) as f:
-        job = pickle.load(f)
+    try:
+        with gzip.open(os.path.join(args.jobdir, gs.JOB_SAVE)) as f:
+            job = pickle.load(f)
+    except FileNotFoundError as e:
+        answer = input('Job does not exists, try to create? (y/N) ').lower()
+        if answer in ('y'):
+            from sisyphus.loader import load_configs
+            from sisyphus import toolkit as tk
+            load_configs()
+            job_list = tk.find_job(os.path.basename(args.jobdir))
+            assert len(job_list) == 1, "none or multiple Jobs were found"
+            tk.setup_job_directory(job_list[0])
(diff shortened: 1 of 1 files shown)


curufinwe:
There is only one problem with this. It only works when you use the default config name. If you have different config names, how will the worker know which config-file to load?

JackTemaki:
> There is only one problem with this. It only works when you use the default config name. If you have different config names, how will the worker know which config-file to load?

The cleaner also does not work when using multiple config files, so I think it is okay to not cover this. I do not see how this can be integrated that easily. I think it is best practice to configure the setup in a way that all configs are loaded when not specifying one.

critias:
I also don't like that it only works if the default config is used.
tk.job_run (https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/toolkit.py#L312) does already something similar. 
How about expanding that function to accept also strings instead of only Jobs?
The function could the be directly called from the system shell using the -c flag, but this is not very convenient. Maybe we should consider add a short cut to access some functions of the toolkit directly.

critias:
The cleaner isn't working with multiple config files? I'm too busy this month to look into it, but that's a bug which should be fixed.

JackTemaki:
> The cleaner isn't working with multiple config files? I'm too busy this month to look into it, but that's a bug which should be fixed.

It does, but you always have to give all configs then, otherwise it will delete Jobs of configs not specified, or was this changed? At least when I tried it a long time ago this was the behavior I think...

curufinwe:
> > The cleaner isn't working with multiple config files? I'm too busy this month to look into it, but that's a bug which should be fixed.
> 
> It does, but you always have to give all configs then, otherwise it will delete Jobs of configs not specified, or was this changed? At least when I tried it a long time ago this was the behavior I think...

And this should be the default behaviour. You could have different python scripts in the root folder, so we cannot reliably determine which python files are sisyphus configs.

critias:
Yes, so cleaning working with multiple configs, but you need to load all configures for sisyphus to know which jobs it should keep. Maybe we should add a warning to remind users about that?

So back to topic: I agree that having this feature would be nice, but it should not only work for the default config.

PR #41, opened 2020-06-18
added error message for parameter bug

JackTemaki:
I had a strange bug, where opening the pickle file would result in the whole config to be loaded again, resulting conflicting emtpy jobs. It took some time until I could find a wrong parameter as cause (it was a wrong class added to a dict which was part of a correct class passed as parameter, the inputs are unfortunately quite complicated sometimes).

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -121,7 +121,11 @@ class JobSingleton(type):
                 job._sis_add_block(b)
 
         # Update alias prefixes
-        job._sis_alias_prefixes.add(gs.ALIAS_AND_OUTPUT_SUBDIR)
+        try:
+            job._sis_alias_prefixes.add(gs.ALIAS_AND_OUTPUT_SUBDIR)
+        except Exception as e:
+            assert False, ("An empty Job was loaded, this can happen if classes were passed as parameters that could"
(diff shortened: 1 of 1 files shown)


critias:
I'm not sure if I understand the problem correctly. Something similar like this failed in your case:

```
class Foo(Job):
  ...

class Bar(Job):
  ...

test = Bar({'test': Foo()})
dump = pickle.dumps(test)
test_loaded = pickle.loads(dump)
```
Correct? The bugfix looks more like treating the symptoms and not the cause, but I need to understand the cause better.

P.S.: Sorry for the late response, I was on vacation.

JackTemaki:
So the issue was that during pickle.load, the whole Sisyphus config (starting from `config/__init__py`) was loaded again, which lead to errors. Unfortunately I corrected everything, so I am not sure if I can produce this error again.

curufinwe (approved):

michelwi:
Could we check for a more specific exception or reraise the original exception together with the explanation. Just so if any other error occured the user is not confused.

JackTemaki:
From my memory I think this was a `'job' as no attribute '_sis_alias_prefixes'`, and one can explicitly catch for that.

PR #43, opened 2020-07-10
make call to h_fsize non hard-coded

christophmluscher:
Current situation: h_fsize in the SGE Engine is hardcoded to 50GB
Change: indroduction of a new key for the requirements dict: "file_size"
Effect: h_fsize can be set by user. The default value remains at 50GB

### DIFF
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -148,16 +148,15 @@ class SonOfGridEngine(EngineBase):
         else:
             out.append('h_rss=%s' % mem)
 
-        if "file_size" in rqmt:
-            try:
-                filesize = "%iG" % math.ceil(float(rqmt['file_size']))
-            except ValueError:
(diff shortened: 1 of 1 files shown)


critias:
You spelled it once file_size and while the variable is called filesize in all other cases. To make it constant with the actual name in the dict I would change it to file_size in all cases. Since you are already using try/except and get you can simply add KeyError to the except case and avoid the first if statement. Also add a comment to make clear that the default value of '50G' can be now be overwritten without any further changes to the code: 
```
            try:
                file_size = "%iG" % math.ceil(float(rqmt['file_size']))
            except (ValueError, KeyError):
                # If a different default value is wanted it can be overwritten by adding
                # 'file_size' to the default_rqmt of this engine.
                file_size = rqmt.get('file_size', "50G")

        out.append('-l')
        out.append('f_fsize=%s' % file_size)
```

christophmluscher:
I incorporated your comment.

PR #48, opened 2021-03-14
add workflow diagramm for the example in the documentation using mermaid

jotix16:
Maybe adding mermaid diagrams for sphinx could improve documentation?

Here I add a diagram for the example. It is still far away from being clear what the example is supposed to do. I spent quite some time and still feel unsure about it.

It feels like the example is forcefully using as many features as possible. While maybe logical, it ends up being very complicated for a novice to get an idea of what sisyphus is supposed to do.

Maybe smaller examples that aim to show only certain features would make more sense. The users can put them together on their own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Sisyphus requires a Python >=3.5 installation with the following additional libr
   Optional to compile documentation:
 
     pip3 install Sphinx
+    pip3 install sphinxcontrib-mermaid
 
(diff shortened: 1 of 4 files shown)


critias:
Thanks, yes the example was build to touch as many things as possible, but I agree that it's not well made to learn Sisyphus. I'm aware of that, but haven't got around yet to redo it.

PR #49, opened 2021-04-30
allow recipe module outside `recipe` package

albertz:
Via: https://github.com/rwth-i6/i6_core/issues/11

Suggested by @JackTemaki .

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -88,8 +88,10 @@ class JobSingleton(type):
         sis_hash = cls._sis_hash_static(parsed_args)
         module_name = cls.__module__
         recipe_prefix = gs.RECIPE_PREFIX + '.'
-        assert module_name.startswith(recipe_prefix)
-        sis_name = os.path.join(module_name[len(recipe_prefix):].replace('.', os.path.sep), cls.__name__)
+        if module_name.startswith(recipe_prefix):
+          sis_name = os.path.join(module_name[len(recipe_prefix):].replace('.', os.path.sep), cls.__name__)
+        else:
+          sis_name = os.path.join(module_name.replace('.', os.path.sep), cls.__name__)
         sis_id = "%s.%s" % (sis_name, sis_hash)
 
         # Update tags


albertz:
Maybe this can be done in some better way. I don't really mind how this is done. But following from the discussion in https://github.com/rwth-i6/i6_core/issues/11, I think it would be nice to support also jobs defined not within the `recipe` package (but e.g. in `i6_core`, which could be outside; and/or maybe other packages).

critias:
The `RECIPE_PREFIX` is used to allow shortening it from the computed hash. This allows renaming the recipe folder if wanted without breaking the hash and shortening the name displayed in the used grid engine. Side note: I don’t think renaming is done in practice since most recipes have `import recipe ..` hardcoded anyway. Loosening this restriction also binds my hands for future code changes in Sisyphus. I already regret not being stricter in some other places.
On the other side, I don’t really see the benefit of loosening this restriction either after reading the discussion you mentioned. If you think `recipe` is too generic as a prefix, just change the `RECIPE_PREFIX` in your `settings.py` and put everything in this folder. If you use multiple git repos, simply add all of them into that folder. Just make sure that all of your recipes follow the convention `from your_unique_prefix.your_repo_name.internal import subfolders`.

Code that is supposed to be used without Sisyphus can be placed in a separate file and imported into the recipes and other tools independent of Sisyphus. Calling recipes files without the intention to use Sisyphus already causes some problems, e.g., you need to make sure Sisyphus is installed in your python environment even though you don’t really need it for that tasks. Keeping them in separate files is cleaner anyway.

albertz:
> The `RECIPE_PREFIX` is used to allow shortening it from the computed hash. This allows renaming the recipe folder if wanted without breaking the hash and shortening the name displayed in the used grid engine.

I don't really understand this argument. Yes, you can rename it now. You can also rename the parent directory. You still cannot rename individual jobs, or anything below. But so what?

And you anyway will not rename the `recipe` package. And also you can not really in practice, as you say yourself.

And you still cannot really mix multiple `recipe` packages together. This is one of the main goals we are currently working on, to have multiple recipe packages which are used together (i6_core, i6_experiments, i6_private, any arbitrary others). So the work around is to keep only this single `recipe` and symlink or git clone them nested in there. But then this `recipe` prefix kind of becomes meaningless.

> Side note: I don’t think renaming is done in practice since most recipes have `import recipe ..` hardcoded anyway.

Yes. There is also no real other way (unless you always do relative imports maybe, which is not standard when you reach a certain depth).

> Loosening this restriction also binds my hands for future code changes in Sisyphus.

Why? How?

> I already regret not being stricter in some other places.

I understand this argument in general, and I also think good and meaningful restrictions can be helpful, or sometimes even mandatory. But then, I think restrictions without real reasons (e.g. every class name must be at least 10 chars long, start with lowercase letter, have 3 upper case letters, and one number) or maybe outdated reasons should also be loosened.

So, the question is not really if restrictions are helpful. The question is if this particular restriction is helpful.

> On the other side, I don’t really see the benefit of loosening this restriction either after reading the discussion you mentioned. If you think `recipe` is too generic as a prefix, just change the `RECIPE_PREFIX` in your `settings.py` and put everything in this folder. If you use multiple git repos, simply add all of them into that folder. Just make sure that all of your recipes follow the convention `from your_unique_prefix.your_repo_name.internal import subfolders`.

There are multiple points:

- We will have multiple repos / packages (i6_core, i6_experiments, i6_private, and potentially other packages) with recipes, which we all want to use together. Changing `RECIPE_PREFIX` is not helpful at all for this. So the argument is, this restriction (feature) of having `RECIPE_PREFIX` doesn't really give us anything, except of making it more inconvenient. And also, in practice you cannot really change it anyway.

- Now all these packages (i6_core etc) depend on this arbitrary magic/prefix. If you would want to use them in other context (e.g. because there is some generic `Lexicon` class, or some generic helper tools, or whatever), you would need to replicate this.

- `recipe` might be a too generic name. This could easily clash with other packages. E.g. [this](https://recipe.readthedocs.io/en/latest/tutorial.html). Or maybe other custom tools. For the same reasons, it's also bad to call a top-level package name `utils` or `sys` or whatever. Imagine if core Python introduces a `recipe` module at some point. E.g. see [other core Python module names](https://github.com/python/cpython/tree/master/Lib). That's why it is often bad to use too generic names. (Ok, maybe this is not so realistic or in practice not really a problem.)

In https://github.com/rwth-i6/i6_core/issues/11 the main argument was for using parts of i6_core, i6_experiments, i6_private also for other purpose and not just exclusively for Sisyphus recipes.

Sure, we can say we have the rule that these repos are exclusively for Sisyphus recipes. Then their names are already bad, and I would have called these repos rather i6_core_sis_recipes, i6_experiments_sis_recipes, etc. And this means we also need further repos i6_core_other or i6_core_common or sth like that, and same for i6_experiments_common, etc, to keep all other stuff which are not Sisyphus recipes. And there surely always will exist such other things. There are already, e.g. [here](https://github.com/rwth-i6/i6_core/tree/main/lib). Or we ignore this, and then

- have duplicate code,
- or having it somehow messy,
- or having it more complex/complicated, and more inconvenient.

That is why I think it is cleaner and simpler to directly use i6_core, i6_experiments etc for both Sisyphus recipes and potentially other utils, code, helpers.

Btw, although it is called i6_core etc, this is probably also helpful for many AppTek things. That's the idea. To share common code, common recipes, common utils, etc.

But also, as we argued, `RECIPE_PREFIX` is not really a feature which can really be used in any way. So it doesn't really seem helpful anyway.

> Code that is supposed to be used without Sisyphus can be placed in a separate file and imported into the recipes and other tools independent of Sisyphus. Calling recipes files without the intention to use Sisyphus already causes some problems, e.g., you need to make sure Sisyphus is installed in your python environment even though you don’t really need it for that tasks. Keeping them in separate files is cleaner anyway.

No, not really. E.g. to import [this file](https://github.com/rwth-i6/i6_core/blob/main/lib/rasr_cache.py), this would work without having Sisyphus in the path. I would argue, any sub packages / modules which do not use functions from Sisyphus should be able to be imported without Sisyphus. You should not have unused imports in your module anyway.

But even if you still get the dependency somehow in there, I don't see this too much being a problem.

critias:
If I understand you right, is the problem simply the `recipe.` prefix in front of all imports. Correct?
I'm bothered by adding all different recipes directly into the main folder. Would it be ok for you to keep all recipes inside the recipe folder and allow to import subfolders inside that folder without the `recipe` prefix? You can stop adding `recipe.` from your imports, and I get to keep a clean main folder.

albertz:
> If I understand you right, is the problem simply the `recipe.` prefix in front of all imports. Correct?

This, but not only. It also makes it inconvenient to reuse such a package because you would need to set it up with such a parent `recipe` root package, and cannot just use it as-is.

> I'm bothered by adding all different recipes directly into the main folder.

But you already have that? You have a `recipe` package directly in your main folder now.

> Would it be ok for you to keep all recipes inside the recipe folder and allow to import subfolders inside that folder without the `recipe` prefix? You can stop adding `recipe.` from your imports, and I get to keep a clean main folder.

Wait, I think we need to differentiate two things here:

- How it is organized on the file system. I.e. where you put your packages, and esp also the recipes. I don't really have strong opinions on this.
   It's fine if this is in the `recipe` sub directory (but maybe it becomes a bit tricky to make it work what you proposed then, while staying compatible at the same time, not sure, I didn't thought this through how this would look like).
   But I actually thought to maybe also put it somewhere else (completely outside) and just set it up in my environment somehow (all the standard ways how you would make some package available for Python).
   My `config.py` would simply do `from i6_core import MyNiceJob` or so. Like I do all other imports as well.

- How it is logically organized in the Python package hierarchy. This is what I'm referring to here in this issue. I want to be able to import jobs from any package, not necessarily the `recipe` package.

critias:
Sorry, I misunderstood the goal of your change at first. The recipes were always inside the main folder. In the beginning, for technical reasons, and this convention stayed after that. So my mind got stuck with the assumption that the recipes will be somewhere inside the main folder if you move them out of the recipe folder.

In principle, I'm fine with placing the recipes somewhere and just importing them via PYTHONPATH. A problem that comes to my mind with this is the possibility to setup Path objects relative to the current module directory e.g. `RelPath = setup_path(__package__);path = RelPath('scripts/awesome_tool.py')`. This will break if the recipes are not in the expected folder. It's probably possible to solve by switching from `__package__` to `__file__` and update `setup_path`, but it needs to be tested. I'll try to look into it this week.

critias:
So after playing around with it a bit I think it's do able, there are probably going to be some limitations e.g. just adding it to PYTHONPATH would not work with a grid engine like sge since the PYTHONPATH is currently not transferred. 
Beside that I still need to put in some work to keep the hash values created by tk.setup_path stable. I will push it here once it's ready.

critias:
Solved with https://github.com/rwth-i6/sisyphus/pull/52

PR #51, opened 2021-05-12
support Enum as Job parameter

JackTemaki:


JackTemaki:
Just as a comment, the reason why `Enum` needs a special treatment is that you will end up in an infinite recursion loop otherwise, as it will be treated as `dict` and then incorrect things happen when you call `.items()` on an `Enum`.

sisyphus/tools.py
```diff
@@ -82,6 +83,8 @@ def extract_paths(args: Any) -> Set:
     elif isinstance(args, (list, tuple, set)):
         for a in args:
             out = out.union(extract_paths(a))
+    elif isinstance(args, enum.Enum):
+        return out
```
christophmluscher, line 87: `Block` and `Enum` have the same operation. Maybe merge?

JackTemaki: would be possible yes, I don't know what is better... saving a line or having potentially more checks...

curufinwe: I'd group by behavior and the behavior for Block and Enum is the same here and I don't see that changing anytime in the future.

PR #52, opened 2021-05-13
Allow more flexibility for recipe loading

critias:
These changes allows to store the recipes without the recipe prefix in arbitrary locations by settings IMPORT_PATHS in settings.py accordingly. e.g.: IMPORT_PATHS = ['config', 'recipe/'] will load everything inside of the recipe/ folder without the recipe prefix.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -165,11 +165,11 @@ MAX_SUBMIT_RETRIES = 3
 #: Default function to hash jobs and objects
 SIS_HASH = sisyphus.hash.short_hash
 
-#: Path to the config directory, not including the directory name 'config'
-CONFIG_PATH = '.'
-
-#: Path to the recipe directory, not including the directory name 'recipe'
-RECIPE_PATH = '.'
+#: List of paths searched for loading config and recipe files. The module name should be part of the path e.g.:
+#: adding 'config' will cause Sisyphus to the current directory for a folder named config to load modules starting
+#: with config, other python files in the current directory will be ignored.
+#: If the path ends with '/' everything inside it will be loaded, similar to adding it to PYTHONPATH.
+IMPORT_PATHS = ['config', 'recipe']
 
 #: The work directory
 WORK_DIR = 'work'
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -88,8 +88,11 @@ class JobSingleton(type):
         sis_hash = cls._sis_hash_static(parsed_args)
         module_name = cls.__module__
         recipe_prefix = gs.RECIPE_PREFIX + '.'
-        assert module_name.startswith(recipe_prefix)
-        sis_name = os.path.join(module_name[len(recipe_prefix):].replace('.', os.path.sep), cls.__name__)
+        if module_name.startswith(recipe_prefix):
+            sis_name = module_name[len(recipe_prefix):]
+        else:
+            sis_name = module_name
+        sis_name = os.path.join(sis_name.replace('.', os.path.sep), cls.__name__)
         sis_id = "%s.%s" % (sis_name, sis_hash)
 
         # Update tags
--- a/sisyphus/loader.py
+++ b/sisyphus/loader.py
@@ -210,14 +210,24 @@ class ConfigManager:
 class RecipeFinder:
     @classmethod
     def find_spec(cls, fullname, path, target=None):
-        for rprefix, rdir in ((gs.RECIPE_PREFIX, gs.RECIPE_PATH), (gs.CONFIG_PREFIX, gs.CONFIG_PATH)):
-            if fullname.startswith(rprefix):
+        for load_path in gs.IMPORT_PATHS:
+            if load_path.endswith(os.path.sep):
+                module_dir = load_path[:-1]
+                module_prefix = ''
+            else:
+                module_dir = os.path.dirname(load_path)
+                if not module_dir:
+                    module_dir = '.'
+                module_prefix = os.path.basename(load_path)
+
+            if not module_prefix or fullname == module_prefix or fullname.startswith(module_prefix + '.'):
                 if path is None:
-                    path = [os.path.abspath(rdir)]
+                    search_path = os.path.abspath(module_dir)
                 elif isinstance(path, str):
-                    path = [os.path.abspath(os.path.join(rdir, path))]
-                spec = PathFinder.find_spec(fullname, path, target)
-                return spec
+                    search_path = os.path.abspath(os.path.join(module_dir, path))
+                spec = PathFinder.find_spec(fullname, [search_path], target)
+                if spec:
+                    return spec
 
     @classmethod
     def invalidate_caches(cls):
--- a/sisyphus/loader.py
+++ b/sisyphus/loader.py
@@ -222,10 +222,12 @@ class RecipeFinder:
 
             if not module_prefix or fullname == module_prefix or fullname.startswith(module_prefix + '.'):
                 if path is None:
-                    search_path = os.path.abspath(module_dir)
+                    search_path = [os.path.abspath(module_dir)]
                 elif isinstance(path, str):
-                    search_path = os.path.abspath(os.path.join(module_dir, path))
-                spec = PathFinder.find_spec(fullname, [search_path], target)
+                    search_path = [os.path.abspath(os.path.join(module_dir, path))]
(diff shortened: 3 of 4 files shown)


JackTemaki:
My current setup crashes with this branch and the following stacktrace (only the relevant part):
```python
~/experiments/tts_asr_2021/recipe/i6_core/corpus/convert.py in <module>
      7 
      8 from recipe.i6_core.lib import corpus
----> 9 from recipe.i6_core.util import uopen
        global recipe.i6_core.util = undefined
        global uopen = undefined
     10 
     11 Path = setup_path(__package__)

~/experiments/tts_asr_2021/recipe/i6_core/util.py in <module>
      9 from sisyphus import *
     10 
---> 11 Path = setup_path(__package__)
        global Path = <class 'sisyphus.job_path.Path'>
        global setup_path = <function setup_path at 0x7f26ae3f89d8>
        global __package__ = 'recipe.i6_core'
     12 Variable = tk.Variable
     13 

~/src/sisyphus/sisyphus/toolkit.py in setup_path(package='recipe.i6_core')
    226     hash_overwrite = package.replace('.', '/')
    227     module = importlib.import_module(package)
--> 228     path = os.path.dirname(module.__file__)
        path = undefined
        global os.path.dirname = <function dirname at 0x7f26b40481e0>
        module.__file__ = None
    229     return RelPath(path, hash_overwrite=hash_overwrite)
    230 

/work/tools/asr/python/3.7.1/lib/python3.7/posixpath.py in dirname(p=None)
    154 def dirname(p):
    155     """Returns the directory component of a pathname"""
--> 156     p = os.fspath(p)
        p = None
        global os.fspath = <built-in function fspath>
    157     sep = _get_sep(p)
    158     i = p.rfind(sep) + 1

TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I have no Idea how `module.__file__` can be None now though....

critias:
Can you try if adding an empty __init__.py into the recipe/i6_core folder solves this problem?

critias:
I'll added a warning and the hint to check for `__init__.py`

JackTemaki:
Okay, then next one, this is what I tried:
 - add a folder "other_recipes"
 - add a "recipe" package inside called  "test_recipe" which contains one file for a dummy job and one file for a helper function, which is imported via `from test_recipe.helper`
- add "other_recipes/" to import path

What I got is:
```python
  File "/u/rossenbach/src/sisyphus/sis", line 14, in <module> 
    main() 
  File "/u/rossenbach/src/sisyphus/sisyphus/__main__.py", line 181, in main 
    from IPython.core import ultratb 
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/IPython/__init__.py", line 55, in <module> 
    from .terminal.embed import embed
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/IPython/terminal/embed.py", line 16, in <module>
    from IPython.terminal.interactiveshell import TerminalInteractiveShell
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 18, in <module>
    from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/__init__.py", line 17, in <module>
    from .application import Application
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/application/__init__.py", line 2, in <module>
    from .application import Application
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/application/application.py", line 3, in <module>
    from prompt_toolkit.buffer import Buffer
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/buffer.py", line 7, in <module>
    from .application.current import get_app
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/application/current.py", line 2, in <module>
    from prompt_toolkit.eventloop.context import TaskLocal, TaskLocalNotSetError
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/eventloop/__init__.py", line 4, in <module>
    from .coroutine import From, Return, ensure_future
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/eventloop/coroutine.py", line 3, in <module>
    from prompt_toolkit.eventloop.defaults import get_event_loop
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/eventloop/defaults.py", line 2, in <module>
    from prompt_toolkit.utils import is_windows
  File "/u/rossenbach/opt/sisyphus/lib/python3.7/site-packages/prompt_toolkit/utils.py", line 12, in <module>
    from six.moves import range
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 963, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 906, in _find_spec
  File "/u/rossenbach/src/sisyphus/sisyphus/loader.py", line 228, in find_spec
    spec = PathFinder.find_spec(fullname, [search_path], target)
UnboundLocalError: local variable 'search_path' referenced before assignment
```

albertz:
Why does this need the custom `RecipeFinder` at all? And what is the purpose of `IMPORT_PATHS`? Why invent sth new which Python itself already provides (`sys.path`)?

JackTemaki:
`IMPORT_PATHS` is just a settings variable to add stuff to the path, of course you could just call `sys.path` directly in your `settings.py`. The RecipeFinder is not really needed. I think the functionality to add package names and not search paths (the distinction of  "recipe" vs. "recipe/") is not really needed, and this is the only addition that the RecipeFinder gives now.

My recommendation would instead of: 
```python
if RecipeFinder not in sys.meta_path:
    sys.meta_path.append(RecipeFinder)
``` 
something like:
```python3
sys.path.append(os.path.abspath("."))  # to not break old setups, also some of my setups did not work without using abspath
sys.path += gs.IMPORT_PATHS
```
At least this worked for my test cases

albertz:
> `IMPORT_PATHS` is just a settings variable to add stuff to the path,
> of course you could just call `sys.path` directly in your `settings.py`.

> The `RecipeFinder` is not really needed.

Yes, this is what I suspected. So then it probably is better to not add such logic to `RecipeFinder`, to keep things simple, right?

> I think the functionality to add package names and not search paths (the distinction of "recipe" vs. "recipe/") is not needed, and this is the only addition that the RecipeFinder gives now.

> ```python
> sys.path.append(os.path.abspath("."))  # to not break old setups, also some of my setups did not work without using abspath
> sys.path += gs.IMPORT_PATHS
> ```

Yes. So do we need `IMPORT_PATHS` at all then? I would vote to keep things simple. Esp if sth doesn't really add any new functionality.

(Btw, that issue with `abspath` sounds a bit strange. But maybe that should be discussed in a separate issue. I would not expect that you need this, or if you do, there is some issue somewhere.)

(Also, side comment, when you modify `sys.path`, I think it is generally recommended to add entries at the front, i.e. like `sys.path.insert(0, ...)` or `sys.path[:0] = ...`.)

critias:
I kept using RecipeFinder to make it possible to keep the exact behavior as before to only import recipe and config from the current folder and not everything in that folder. I'm not sure if this is really needed though. If you put a module in your main folder that collides with a module name that you import it's kinda your own fault.
So if the consent here is just to import the whole folder, I'll change that.

IMPORT_PATHS is mainly used to ensure that jobs that are submitted to a grid engine will know about these paths. The current value of PYTHONPATH would not necessarily be transmitted to the worker otherwise.

The abspath thing is strange. I can't reproduce it with my toy workflow.

About the UnboundLocalError: Did you try to import a module that doesn't exist? That case isn't handled correctly with RecipeFinder, as I just found out trying to reproduce your error.

albertz:
> I kept using RecipeFinder to make it possible to keep the exact behavior as before to only import recipe and config from the current folder and not everything in that folder.

Yes, sure, I did not mean to remove `RecipeFinder`. I just meant that I don't see a need in adding your change to it, i.e. in adding complexity, because Python itself already does exactly the same thing.

JackTemaki (approved):
This works fine now for the old setups, and for the new "test" setup as well.

PR #54, opened 2021-06-01
rqmt fix for Task

JackTemaki:


critias:
Is there a reason you don't use simply `self._rqmt = rqmt.copy()`?

JackTemaki:
> Is there a reason you don't use simply `self._rqmt = rqmt.copy()`?

Is not thinking a valid reason? :D I thought about deepcopy first, and then changed it to this... I will update this tomorrow.

JackTemaki:
Sorry I completely forgot about this!

PR #56, opened 2021-06-12
Add marker IS_WORKER to global settings

critias:


JackTemaki:
I think this will work, but I am not sure if putting it into `global_settings.py` without a comment is a good idea, because people might think that this is a normal settings parameter....

curufinwe:
I agree with @JackTemaki . Maybe this can be a property of the job, i.e. `sis_in_worker`.

JackTemaki:
In the Job itself is maybe not a good idea, because I want to use this also outside of jobs... But I guess if this is a static variable in the Job class there can be some "helper" function to call it...

critias:
You are right global_settings.py was a bad place, but I don't like to attach it to the job class. The toolkit looks like a good place to me.

JackTemaki (approved):

JackTemaki:
Ah I was too fast, the `global_settings.py` edit is still in there, which needs to be removed...

sisyphus/global_settings.py
```diff
@@ -354,6 +354,8 @@ def update_global_settings_from_file(filename):
     update_global_settings_from_text(content, filename)
 
 
+IS_WORKER = False
```
JackTemaki, line 357: This needs to be removed

JackTemaki (changes requested):

sisyphus/worker.py
```diff
@@ -153,6 +154,8 @@ def stop(self):
 def worker(args):
     # Change job into error state in case of any exception
     gs.active_engine = gs.engine().get_used_engine(args.engine)
+    sisyphus.toolkit._sis_running_in_worker = True
```
curufinwe, line 157: Technically I guess we should also set this back to false in a finally block afterwards.

critias: I don't think it's relevant at the moment, but you are right. I added it.

JackTemaki (approved):

PR #59, opened 2021-06-28
Added fallback and rformat to Delayed objects

critias:
Added two new methods to Variables:
- fallback: If Variable is not set yet the given default value will be returned
- rformat: Reversed format, a.rformat(b) == b.format(a), useful as shortcut to created a DelayedFormat object.

Also is_set is too propagated backwards through a chain of DelayedBase Objects.

### DIFF
--- a/sisyphus/delayed_ops.py
+++ b/sisyphus/delayed_ops.py
@@ -3,9 +3,19 @@ from sisyphus.tools import try_get
 
 
 class DelayedBase:
-    def __init__(self, a, b):
+    def __init__(self, a, b=None):
         self.a = a
         self.b = b
+
+    @staticmethod
+    def _is_set_helper(var):
+        if isinstance(var, DelayedBase):
+            if not var.is_set():
+                return False
+        return True
+
+    def is_set(self):
+        return self._is_set_helper(self.a) and self._is_set_helper(self.b)
 
     def __add__(self, other):
         return DelayedAdd(self, other)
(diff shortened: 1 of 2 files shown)


JackTemaki:
Just out of curiosity, what would be an application scenario for the DelayedFallback? I can not image one right now...

critias:
Mainly for reports which can contain unfinished variables. e.g.:

`tk.register_report('status_report', 
{'test1': var1, 'test2': var2}, 'Test1: {test1}\nTest2: {test2}')`

If var1 and var2 are normal Variables it would returns a fairly long string telling you that this Variable is not finished which looks ugly in a larger report. Some thing like `'{test2:.2f}'` also wouldn't work since Variable can not be easily formatted. With this change it's possible to do something like:

`tk.register_report('status_report', {'test1': var1.rformat('{:.2f}'.fallback('?'), 'test2': var2.rformat('{:.2f}').fallback('?')}, 'Test1: {test1}\nTest2: {test2}')`

Resulting in nicely shortened floats and simply '?' if the Variable isn't finished yet.

sisyphus/delayed_ops.py
```diff
@@ -43,12 +53,21 @@ def __getitem__(self, key):
...
+        a.rformat(b, *args, **kwargs) is mapped to b.format(a, *args, **kwargs) """
+        return DelayedFormat(fstring, self, *args, **kwargs)
+
     def replace(self, *args, **kwargs):
         return DelayedReplace(self, *args, **kwargs)
 
     def function(self, func, *args, **kwargs):
         return DelayedFunction(self, func, *args, **kwargs)
 
+    def fallback(self, fallback):
```
JackTemaki, line 67: Should the fallback then not be limited to "Non-Delayed" types?

critias: Well I guess it could be possible to stack these fallback methods, even though I don't really see a use case for it...
You are right, I either need to restrict it to Non delayed types or add try_get(...) around self.b inside DelayedFallback.get
Since I don't see much danger in allowing this usage I think we can allow it as long as it doesn't cause any problems.

JackTemaki (approved):

PR #60, opened 2021-06-28
User settings file

critias:
Adding support for a user base settings file stored under `~/sisyphus_settings.py`
Just checking if you see any reason not to add it.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -347,7 +347,6 @@ def update_global_settings_from_file(filename):
     :param str filename:
     :return: nothing
     """
-    # skip if settings file doesn't exist
     globals()['GLOBAL_SETTINGS_FILE'] = filename
 
     content = ''
@@ -355,6 +354,7 @@ def update_global_settings_from_file(filename):
         with open(filename, encoding='utf-8') as f:
             content = f.read()
     except IOError as e:
+        # skip if settings file doesn't exist
         if e.errno != 2:
             raise e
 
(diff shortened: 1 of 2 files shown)


sisyphus/global_settings.py
```diff
@@ -392,5 +392,6 @@ def update_global_settings_from_env():
 # Set to False to disable Warning of unset engine
 ENGINE_NOT_SETUP_WARNING = True
 
-update_global_settings_from_env()
+update_global_settings_from_file(GLOBAL_SETTINGS_FILE_USER)
 update_global_settings_from_file(GLOBAL_SETTINGS_FILE_DEFAULT)
+update_global_settings_from_env()
```
JackTemaki, line 397: This order change might have an effect on some setups that make use of setting global settings from env, but I think most people are not doing it.

JackTemaki (approved):

sisyphus/global_settings.py
critias: I think it was an accident that env was before file. The main use cases of env is to overwrite settings without changing the file, e.g. testing something in a directory that doesn't belong to you.

curufinwe:
I don't like this change, because now you need again to chase user settings files to reproduce a setup. At the moment you just need to look at the setup folder to know how to run the setup.

critias:
I forgot that you put a lot more stuff into your settings.py. It's kinda inconvenient for me to basically copy the same settings.py from experiment folder to experiment folder, but I see that this would motivate bad practice in your case.

PR #61, opened 2021-06-28
add worker check for Delayed

JackTemaki:
I thought it is a good idea to add a gs variable that prohibits resolving any "Delayed" object  while still being in the "graph" thread. I saw that this is the most common error when working with Sisyphus.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -306,6 +306,8 @@ WARNING_NO_FUNCTION_CALLED = True
 #: Warn if an absolute path inside the current directory is created
 WARNING_ABSPATH = True
 
+# Prohibit resolving paths in graph construction
+DELAYED_CHECK_FOR_WORKER = False
 
 # Internal functions
 def update_global_settings_from_text(text, filename):
--- a/sisyphus/job_path.py
+++ b/sisyphus/job_path.py
@@ -4,11 +4,27 @@ import os
 import logging
 import gzip
 import pickle
+from functools import wraps
 
 import sisyphus.tools as tools
 import sisyphus.global_settings as gs
 from sisyphus.delayed_ops import DelayedBase
 from sisyphus.tools import finished_results_cache
+
+
+def check_is_worker(get_func):
+    """
+    Prohibits calling a get function while being in the "graph" thread, thus
+    causing bugs/inconsistent behavior in Sisyphus.
+    """
+    @wraps(get_func)
+    def check(*args, **kwargs):
+        if gs.DELAYED_CHECK_FOR_WORKER:
+            from sisyphus.toolkit import running_in_worker
+            assert running_in_worker()
+        return get_func(*args, **kwargs)
+
+    return check
 
 
 class Path(DelayedBase):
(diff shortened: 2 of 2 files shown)


curufinwe (approved):
Good idea!

critias:
This would break a lot of our code and I consider it ok to resolve Variable inside the graph especially in combination with async workflows.

I have to admit I regret allowing to converting a Path and Variable to the Path value by overwriting __repr__. While this resulted in nice code in some places it makes the debugging ofter harder than it should be since it's complicated to tell Path and Variable apart from regular str/int/float etc. 

I had some ideas to get this fixed, but haven't tried any of it since nobody complained so far and changing the behavior of a very basic part of Sisyphus has a high likelihood of breaking stuff. If I'm not the only one bothered by it I'll put some thought into finding a solution for it.

JackTemaki:
> This would break a lot of our code and I consider it ok to resolve Variable inside the graph especially in combination with async workflows.

I already assumed that this could happen, this is why I added a global option for this to disable this. Do you think it is still a bad idea to add this? And maybe you can explain how async workflows work, I have absolute no idea about that.

critias:
The idea of async workflows is to compute parts of the graph before continue reading the rest of the workflow e.g.:
```
async def workflow():
    # Some job outputs
    a = SomeJob('some_input').out
    b = SomeJob('other_input').out
    
    # Tell sisyphus to wait until a and b are finished
    await tk.async_run([a, b])
    
    # You can now do something depending on the output of a or b
    if a.get() > b.get():
        c = a
    else:
        c = b
```

So it needs to be possible to resolve a and b. Throwing an error as default behavior is a bad idea at this point. Beside that checking it in Variable.get and Path.get should be enough. All delayed operations either end with one of these or should always produce the same result.

I also don't like to start treating async and non async workflows differently. 

I think the problems with converting it inside the graph is when someone does it implicitly at some point. e.g. converting it in a line like: `"Foo bar %s" % var`. Do you agree? If not what is the main problem in your experience?

JackTemaki:
> So it needs to be possible to resolve a and b

But this looks like that `async_run` can just add a "flag" in a and b that it is okay to access them.

> If not what is the main problem in your experience?

People accidentally convert it into a string, and then use it as a parameter to a job, resulting in completely incorrect hashes. But still you get no notification that you are doing something wrong.

curufinwe:
One more use-case to consider: what about reports? Are they treated as if they run in a worker? Maybe the distinction here should not be worker / not worker, but initial graph loading / rest?

critias:
You are right, this will probably break the reports as well. I'm not so happy with this as a whole, but accepted it as a debugging tool for you guys. Making a distinction between initial loading and the rest could again cause problems with the update function.

How about this:
- I'll add a flag to raise an exception if an unfinished Variable is resolved this should already catch many cases. The main problem is if someone is working on a partial finished graph, in this case it would not raise an Exception and the problem become only appeared when used with in a new workflow or emptied work directory. 
- I'll add another flag to change the __str__ and __repr__ behavior of Path and Variable to something like '<Path ....>'. This should break most things where path is automatically converted and forces people to use .get().

Do you think this would help to catch most of you problems?

PR #62, opened 2021-07-02
WIP: Check size of input files to handle nfs sync issues

mirkovogel:


curufinwe:
Note that the finished file is also put in the finished.tar.gz file at job cleanup.

sisyphus/job.py
```diff
@@ -651,9 +655,80 @@ def _sis_all_path_available(self):
...
+    def _sis_get_file_stats(self):
+        """
+        Returns a triple for every file below `work` and `output`: path, modification time, size.
+
+        These stats are written to the `usage` files by the `LoggingThread`, and read by
+        Job._sis_get_expected_file_sizes.
+
+        """
+        stats = []
+        below_work = pathlib.Path(self._sis_path(gs.WORK_DIR)).rglob("*")
```
critias, line 667: I think this runs directly before switching back from the job work directory to the experiment directory. This will fail from the job work directory since self._sis_path returns a relative path starting from the experiment directory.

sisyphus/task.py
```diff
@@ -209,6 +215,58 @@ def run(self, task_id, resume_job=False, logging_thread=None):
...
+
+        # Make sure the files have the required size
+        logging.info("Waiting for the filesystem to sync files ...")
+        for path, expected_size in expected_sizes:
+            logging.debug(path)
+
+            start = time.time()
+            while True:
+                try:
+                    # Use absolute path?
```
critias, line 254: This run in the experiment directory, so relativ paths should work.

critias (commented):
The overall approach looks good to me, let's see if it works as expected.

mirkovogel:
Side note: This change does not break old setups. If no size info is available for a given job / task, no checks are run.

sisyphus/global_settings.py
```diff
@@ -210,6 +210,10 @@ def file_caching(path):
 WAIT_PERIOD_JOB_CLEANUP = 10
 #: How many seconds should all inputs be available before starting a job to avoid file system synchronization problems
 WAIT_PERIOD_MTIME_OF_INPUTS = 60
```
critias, line 212: WAIT_PERIOD_MTIME_OF_INPUTS can be removed since the check inside of task is removed. Setting it should also be removed in toolkit.setup_script_mode

sisyphus/job.py
```diff
@@ -651,9 +656,96 @@ def _sis_all_path_available(self):
...
+        If `task` is given, only usage files from these tasks are read.
+
+        """
+        # The job might be a Job object or the job directory
+        try:
+            job_dir = job._sis_path()
+        except AttributeError:
+            job_dir = job
+
+        if os.path.exists(os.path.join(job_dir, gs.JOB_FINISHED_ARCHIVE)):
```
critias, line 706: This could alternatively also be checked doing something like `subprocess.check_output(['tar', '-Oxf', os.path.join(job_dir, gs.JOB_FINISHED_ARCHIVE), 'usage.run.1'])`, but I'm not sure if this would be worth the effort. The given solution should work in nearly all cases if the clean up timeout is large enough.

sisyphus/task.py
```diff
@@ -209,6 +215,64 @@ def run(self, task_id, resume_job=False, logging_thread=None):
...
+            for i in self._job._sis_inputs:
+                if not i.creator:
+                    logging.info("Cannot check the size of %s, it's not created by sisyphus.", i)
+                    continue
+
+                other_job_sizes = job.Job._sis_get_expected_file_sizes(i.creator)
+                # If the job has been cleaned up, no size info is available, but we can safely
+                # assume that enough time has passed so that all files are synced.
+                if other_job_sizes:
+                    expected_sizes[i.rel_path()] = other_job_sizes[i.rel_path()]
```
critias, line 248: This fails if a path is only used as prefix without representing a real file. We could change it to something like this:
```
try:
    expected_sizes[rel_path] = other_job_sizes[rel_path]
except KeyError:
    for k, v in other_job_sizes.items():
        if k.startswith(rel_path):
            expected_sizes[k] = v
```
This could also be used if the path is pointing to a directory.

sisyphus/job.py
```diff
@@ -651,9 +656,96 @@ def _sis_all_path_available(self):
...
+                        logging.warning("%s contains no file_stats (was created by an older version of sisyphus).")
+                        stats = []
+                        break
+                
+                if stats:
+                    break
+                if time.time() - start > timeout:
+                    logging.error("%s not synced for more than %ds, file_stats still empty.", fn, timeout)
+                    raise TimeoutError
+                logging.info("%s not synced yet, file_stats still empty.", fn)
```
critias, line 731: This will cause problems if a path is available and should be used before the job is finished e.g. training of a neural model.
We could require that these paths have a special attribute set. They could then be excluded from this check.

sisyphus/job.py
```diff
@@ -651,9 +656,96 @@ def _sis_all_path_available(self):
...
+        m_times = defaultdict(int)
+        sizes = dict()
+        
+        exp = "{0}.{1}.*".format(gs.PLOGGING_FILE, task if task else "*")
+        for fn in pathlib.Path(job_dir).glob(exp):
+            start = time.time()
+            while True:
+                with open(fn) as f:
+                    try:
+                        stats = literal_eval(f.read())["file_stats"]
```
critias, line 719: It can happen that this raises a SyntaxError if the file is accessed while it's being written.

sisyphus/task.py
```diff
@@ -112,22 +122,18 @@ def run(self, task_id, resume_job=False, logging_thread=None):
...
-            # if an input file is too young it's may not synced in a network filesystem yet
-            try:
-                input_age = time.time() - os.stat(i.get_path()).st_mtime
-                time.sleep(max(0, gs.WAIT_PERIOD_MTIME_OF_INPUTS - input_age))
-            except FileNotFoundError:
-                logging.warning('Input path does not exist: %s' % i.get_path())
+        logging.info("Inputs:\n%s", "\n".join( str(i) for i in self._job._sis_inputs))
+        
+        try:
+            self._wait_for_input_to_sync()
```
critias, line 132: It would be could to be able to switch off the new check with an entry in the setting file if something goes wrong. Even better: switch back to the old timeout.

sisyphus/task.py
```diff
@@ -112,22 +122,18 @@ def run(self, task_id, resume_job=False, logging_thread=None):
...
+        except TimeoutError:
+            self.error(task_id, True)
```
critias, line 134: This doesn't really stop the task from running. It's sets the error marker and then continues. Once the task is finished a finished marker is set and the error marker is ignored.

mirkovogel:
@critias : I wanted to get this PR merged before "vacation" (=no kindergarden) starts, which didn't happen because I spent last week in bed, not in front of a computer screen. As I won't be able to do so until 8/13, I invite you to take over this PR. :-)

mirkovogel:
@critias I got sidetracked for quite some time ... How about the current situation of the cluster? Is it running so smoothly that this PR has become obsolete? Starting next week I'd have time to implement the changes you suggested.

critias:
I continued working on it here: https://github.com/rwth-i6/sisyphus/tree/check-output-size
It seems to be working ok for me, but I got also sidetracked since the overall situation got better. Let's have a call discuss how to continue from here.

PR #63, opened 2021-07-05
Variable not set

critias:
Added setting to make accessing of Variables and Paths more restrictive.

### DIFF
--- a/sisyphus/job_path.py
+++ b/sisyphus/job_path.py
@@ -196,7 +196,10 @@ class Path(DelayedBase):
         return self.get_cached_path()
 
     def __repr__(self):
-        return repr(str(self))
+        if gs.LEGACY_PATH_CONVERSION:
+            return repr(str(self))
+        else:
+            return '<Path %s>' % self.get_path()
 
     def __lt__(self, other):
         """
@@ -413,4 +416,10 @@ class Variable(Path):
         return str(self.get())
 
     def __repr__(self):
-        return str(self.get())
+        if gs.LEGACY_VARIABLE_CONVERSION:
+            return repr(self.get())
+        else:
+            value = ''
+            if self.is_set():
+                value = ' %s' % self.get()
+            return "<Variable %s%s>" % (self.rel_path(), value)
--- a/sisyphus/job_path.py
+++ b/sisyphus/job_path.py
@@ -27,6 +27,11 @@ def check_is_worker(get_func):
     return check
 
 
+class VariableNotSet(Exception):
+    """Variable is not set"""
+    pass
+
+
 class Path(DelayedBase):
     """
     Object do hold the connecting path to files:
@@ -363,6 +368,8 @@ class Variable(Path):
                                                           < gs.CACHE_FINISHED_RESULTS_MAX_SIZE))
     def get(self):
         if not self.is_set():
+            if gs.RAISE_VARIABLE_NOT_SET_EXCEPTION:
+                raise VariableNotSet("Variable is not set (%s)" % self.get_path())
             if self.backup is None:
                 return "<UNFINISHED VARIABLE: %s>" % self.get_path()
             else:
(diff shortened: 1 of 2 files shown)


sisyphus/job_path.py
```diff
@@ -363,6 +374,8 @@ def is_set(self):
                                                           < gs.CACHE_FINISHED_RESULTS_MAX_SIZE))
     def get(self):
         if not self.is_set():
+            if gs.RAISE_VARIABLE_NOT_SET_EXCEPTION:
```
curufinwe, line 377: Why is it an error to access a variable that has a backup?

curufinwe (commented):
How about only changing __repr__ and not __str__ ? Would fit more with https://docs.python.org/3/reference/datamodel.html#object.__str__ and https://docs.python.org/3/reference/datamodel.html#object.__repr__

sisyphus/job_path.py
critias: I don't really like the solution of having a backup value at this point since it doesn't allow to return `None`. The recently add `fallback` method solves this in a cleaner way. So I plan to remove the possibility to add a backup directly to the Variable at some point.

critias:
Converting Path into strings caused frequently problems of losing the creator inside the graph. So I hoped that forcing people to use `get` would make sure that it doesn't happen by accident, but it also hurts the usability a lot.

So I guess you are right, keeping `__str__` and changing only `__repr__` would probably be the better way to go here.

sisyphus/job_path.py
curufinwe: In the ASR recipes it's not used. Is backup used in any of the MT recipes? If not you could remove it directly.

critias: It's used inside the MT recipes, so I would like to keep both for a while for backwards compatibility during the conversion.

sisyphus/job_path.py
```diff
@@ -403,7 +413,16 @@ def __fs_get__(self, step):
...
 
     def __str__(self):
-        return str(self.get())
+        if gs.FAST_VARIABLE_CONVERSION:
+            return str(self.get())
+        else:
+            value = ''
+            if self.is_set():
+                value = ' %s' % self.get()
+            return "<Variable %s%s>" % (self.rel_path(), value)
```
curufinwe, line 422: I think this is a mixup, this logic should go to __repr__ and the logic from __repr__ should go here

sisyphus/job_path.py
```diff
@@ -403,7 +413,16 @@ def __fs_get__(self, step):
...
 
     def __repr__(self):
-        return str(self.get())
+        if gs.FAST_VARIABLE_CONVERSION:
```
curufinwe, line 425: thish should be in __str__, but for __str__ we dont really need to change it anyomer (same as Path).

curufinwe (changes requested):

JackTemaki:
I am not sure I understand what the changes actually intend to do... can you explain this in a few words? (this "fast" conversion)

critias:
The fast conversion is the current behavior of Sisyphus which I would like to discourage in the future. Especially overwriting __repr__ tends to make debugging more problematic and I would like to disable this by default.
I'm not sure about the __str__ overwrite, maybe it's better to keep the old behavior for this. Any thoughts from your side?
I thinking about rename the settings to LEGACY_PATH_CONVERSION and LEGACY_VARIABLE_CONVERSION. Which name would you consider more descriptive?

JackTemaki:
LEGACY_... sounds much better than FAST_...

curufinwe:
The str behavior is probably used in some places of the ASR recipes, but I'm not 100% sure. So I'd like to keep that one "as-is".

critias:
So the current version should be ok, right?

curufinwe (approved):

PR #68, opened 2021-10-11
add DelayedSlice

JackTemaki:


critias:
LGTM, you  could also try to create `DelayedSlice` automatically if `__get__` gets a slice object as key: https://docs.python.org/3/library/functions.html#slice

PR #71, opened 2021-12-09
Apply __add__ and __getitem__ directly on Path

critias:
A possible solution to:
https://github.com/rwth-i6/i6_core/issues/198

Todo: I need to split the common functionality between Path and Variable into a new class, instead of having Variable as a subclass of Path to clean up the code a bit.

### DIFF
--- a/example/config/__init__.py
+++ b/example/config/__init__.py
@@ -18,7 +18,8 @@ def main():
 
 
 async def async_main():
-    input_data = tk.Path('data/5lines.txt', tags={'5lines'})
+    input_data = tk.Path('data', tags={'5lines'})
+    input_data = input_data.join_right('5lines.txt')
     spliter = parallel.LineSpliter(input_data)
     # Test if requesting gpu works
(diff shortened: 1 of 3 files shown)


critias:
This test failes: https://github.com/rwth-i6/sisyphus/runs/4469494237?check_suite_focus=true
The question here is if this operation is really useful for a Path object...

curufinwe:
The problem is that the previous implementation of `__getitem__` is a wrapper object that will call `get()` first (which makes the path an absolute path) and then apply the slice. With the new implementation we do the slicing on the path first and then return a new path object (which, because we slice only a single character) will be a relative `Path` object and when calling `get()` on the new `Path` object we get the sliced path + `gs.BASE_DIR`.

If we want to keep the invariant that delayed_operation(path).get() should be the same as operation(path.get()) we should probably not change `__getitem__`. But I see no problem just adding new member functions that manipulate the Path object explicitly, e.g.
```
def join_right(self, extension):
  new = self.copy()
  new.path = os.path.join(new.path, extension)
```

The __add__ method can probably stay the same.

critias:
Done

curufinwe (approved):

PR #75, opened 2022-02-24
Bugfix for get_job_from_path

dthulke:
The path argument in get_job_from_path is overwritten by the variable in the inner loop.
This causes that get_job_from_path always returns None.

Noticed this when trying to migrate a large graph.


critias:
Thanks!

PR #77, opened 2022-03-11
Call check_engine_limits on initial submit

dthulke:
This moves the call of `gs.check_engine_limits` from `Task.update_rqmt` to `EngineBase.get_rqmt`. This way the requirements are already checked on the initial submit and not only after the requirements are updated.

This helps to avoid issues if the initial job requirements are already too high for the current engine.

Also based on the documentation in global_settings.py, I expected that the method would be called also on the first submission of tasks.

### DIFF
--- a/sisyphus/engine.py
+++ b/sisyphus/engine.py
@@ -102,6 +102,9 @@ class EngineBase:
         else:
             # never submitted so far, use default values
             rqmt = default_rqmt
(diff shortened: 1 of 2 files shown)


critias:
Thanks, you are right. It was originally placed in update_rqmt to ensure that sisyphus doesn't automatically increase the requirements above its limits, but it's better to check all requirements.

PR #85, opened 2022-06-20
extract_paths more efficient

albertz:
This gives me some huge speedup, of a big GMM pipeline, from 18 sec runtime down to 12 sec runtime.

### DIFF
--- a/sisyphus/tools.py
+++ b/sisyphus/tools.py
@@ -78,30 +78,36 @@ def extract_paths(args: Any) -> Set:
     :rtype: set
     """
     out = set()
-    if isinstance(args, Block) or isinstance(args, enum.Enum):
-        return out
-    if hasattr(args, '_sis_path') and args._sis_path is True:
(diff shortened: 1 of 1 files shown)


albertz:
> This PR breaks my setups, we have to revert this.

@JackTemaki  Can you give some details?

JackTemaki:
> > This PR breaks my setups, we have to revert this.
> 
> @JackTemaki Can you give some details?

Not yet, just that nearly all of the dependencies are lost. I now have to start debugging.

albertz:
Ah I think I know. This is probably because there are temp `tuple` objects created (via whatever `__sis_state__` or `__getstate__` you have) and then they are deleted again and so we have the same `id(obj)` for different objects...

albertz:
Ok I will try to fix this.

albertz:
@JackTemaki Ok I think I fixed it, see #89.

PR #86, opened 2022-06-20
sis_hash_helper more efficient

albertz:
This gives me some huge speedup, of a big GMM pipeline, from 12 sec runtime down to 2 sec runtime.


albertz:
Hm ok I need to do some debugging, this seems to change the hash.

JackTemaki:
Yes, my setup before the change:
`error(7) runnable(5) waiting(649)`

My setup after the change:
`error(3) runnable(12) waiting(247)`

albertz:
Ok, similar fix as in #89. Now for my simple test I get the same hash.

albertz:
However, the speedup is much less now. Need to do some benchmarks. I think it should still be faster than before though.

albertz:
> However, the speedup is much less now. Need to do some benchmarks. I think it should still be faster than before though.

Actually, no? It seems slightly slower. Maybe due to higher memory consumption? I need to investigate this a bit more...

sisyphus/hash.py
```diff
@@ -69,13 +69,20 @@ def get_object_state(obj):
         return args, state
 
 
-def sis_hash_helper(obj):
+def sis_hash_helper(obj, _visited=None, _add_to_visited=True):
```
critias, line 72: _add_to_visited isn't needed here anymore, right?

sisyphus/hash.py
```diff
@@ -69,13 +69,20 @@ def get_object_state(obj):
...
     :param object obj:
+    :param _visited: (internal use)
+    :param _add_to_visited: (internal use)
     :rtype: bytes
     """
+    if _visited is None:
+        # keep ref to obj alive to avoid having the same id for different objs
+        _visited = {}  # id -> (bytes, obj)
+    if id(obj) in _visited:
+        return _visited[id(obj)][0]
```
critias, line 85: Since we are already tracking the visited nodes we could also throw a meaningful error if a Node is visited again before the hash is computed. This means that there is a circular dependency, I think right now this would simply result in a "RecursionError: maximum recursion depth exceeded" exception.
`_visited[id(obj)] = None` could be used to mark nodes which are no finished yet.

sisyphus/hash.py
```diff
@@ -88,31 +95,35 @@ def sis_hash_helper(obj):
...
     if len(byte_str) > 4096:
         # hash long outputs to avoid arbitrary long return values. 4096 is just
         # picked because it looked good and not optimized,
         # it's most likely not that important.
-        return hashlib.sha256(byte_str).digest()
-    else:
-        return byte_str
+        byte_str = hashlib.sha256(byte_str).digest()
+    if _add_to_visited:
+        _visited[id(obj)] = (byte_str, obj)
```
critias, line 128: `obj` is added here to ensure no id is used twice? If so a short comment would be helpful.

albertz:
> > However, the speedup is much less now. Need to do some benchmarks. I think it should still be faster than before though.
> 
> Actually, no? It seems slightly slower. Maybe due to higher memory consumption? I need to investigate this a bit more...

I'm also thinking about actually do more clever caching. But not exactly sure where. Unfortunately the objects we get here are often `dict` or `list` or some other config objects, and they are all mutable, so you cannot really cache the hash, unless you make sure the cache is correctly invalidated, which is not easily possible always (e.g. no idea how you would do that for a `dict`). Maybe in `JobSingleton.__call__` we can do it for the args, because afterwards the args should anyway not change anymore.

sisyphus/hash.py
albertz: Yes but do we really get into such cases? So far any such case would have resulted in RecursionError, as you said, for this function and many other functions as well. And I think RecursionError is also fine here. I think we don't need to make the logic more complex (and also slightly slower because of the extra checks) just to provide a nicer error message.

sisyphus/hash.py
albertz: There is already a comment on the content of `_visited` above:
```python
# keep ref to obj alive to avoid having the same id for different objs
```
This is not enough?

albertz:
Also, I think the current change in this PR here does not fully captures all the recursive calls of `sis_hash_helper`. The recursive calls are often through `obj._sis_hash` calls.

critias:
> Also, I think the current change in this PR here does not fully captures all the recursive calls of `sis_hash_helper`. The recursive calls are often through `obj._sis_hash` calls.

You could also cache the result of `obj._sis_hash`

PR #88, opened 2022-06-20
Revert "extract_paths more efficient"

JackTemaki:
This reverts commit 962043308aed862464be044213745a73367ee8d7.

#85 caused that jobs with RASR configs do not have any dependencies to previous jobs anymore, causing that all jobs want to run at the same time.


JackTemaki:
Yes, seems to work.

PR #98, opened 2022-08-31
add setup_job_directory shortcut

JackTemaki:


sisyphus/shortcuts.py
```diff
@@ -179,6 +188,12 @@ def remove_job_and_descendants(args):
     call_sis(call)
 
 
+def setup_job_directory(args):
+    for conf_file in args.argv:
```
critias, line 192: Does passing all config files at once cause problems?

critias (changes requested):
If a job isn't inside the graph this will probably crash since `tk.sis_graph.find("%s")` returns an empty array. Given that you check each config file individually it will not work if the first config file doesn't contain the job.

Either pass all config files at once or allow only one config file as input.

curufinwe:
I don't mind having this shortcut. @JackTemaki is this still relevant for you?

JackTemaki:
> I don't mind having this shortcut. @JackTemaki is this still relevant for you?

Well, this is just a small convenience increase that you do not have to open the console for that, I would not call it relevant.

michelwi:
some similar functionality is available now via #239, should we close this?

PR #101, opened 2022-09-02
change keep value remove msg

Atticus1806:
Small PR to change the msg. that shows when using the clenup keep value function.
I found "Keep Value too low" confusing, since to me it suggested that the value I set was too low instead of removing jobs with too low value. 
The new msg. is just a suggestion from me, I am fine with other suggestions if you dont like this one.


critias:
Thanks, that sounds much better.

Issue #91, opened 2022-06-24
Use config file from recipes directly


albertz:
Consider `i6_experiments.users....experiments...`.

In many cases, I don't need to have a separate config file. Or such a config while would basically only do an import, and maybe call some `run()` function and nothing else. However, in principle, I would want to directly do this:
```
sis m recipe/i6_experiments/...
```
But Sisyphus complains that it must be in the `config` directory.

Can we add support for that?

JackTemaki:
In my current setup it is structured like:
`./sis m` -> run all experiments
`./sis m config.active` -> run all experiments I am currently working on
`./sis m config.librispeech.conformer` -> run my librispeech conformer experiments

albertz:
I don't really understand your argumentation. I have config files just like you say, they are just in a different location, namely directly in `i6_experiments`.

Also, I don't understand the argumentation, why it would be bad to add this flexibility? Or if you did not say that, why is it relevant what you said?

Also, I don't understand the argument, why is having overhead a good thing?

JackTemaki:
If you prefer to have this exact structure somewhere else, just symlink it as config folder.

albertz:
A symlink would not work because I use relative imports.

albertz:
> I wont spend time on this but feel free to do so

@christophmluscher I thought you actually also use it exactly this way, or not?

christophmluscher:
no I use symlinks. but then I dont use relativ imports. I dont like `./sis m recipe/i6_experiments/users/luescher/configs/lilbrispeech/...`, I prefer `./sis  m config/...`

critias:
The check was added since there used to be an internal reason that config and recipes had to be separated. I don't know of any technical reason for that anymore in the current code base. The cleaner also shouldn't care where the config file is located once the graph is loaded.

That said: I don't think it's a good idea to put your config files into the recipe folder, but I also don't think that it's my decision to make which convention you should follow.

You could just remove the assertion and try it.

PR #102, opened 2022-09-09
config from recipe, maybe remove import path prefix

albertz:
Extension to #91, #100.

E.g. when you now do:
```
sis m recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py
```
When having `"recipe/"` in my `gs.IMPORT_PATHS`, I want that it loads the module `i6_experiments.users.zeyer.experiments.exp2022_07_21_transducer.pipeline_swb_2020`, not with the `recipe.` prefix.

To be fair, it's a bit ambiguous when you have both `"recipe/"` and `"recipe"`. Then both work but then it depends on how the remaining recipes are implemented, and whether their imports are using the `recipe.` prefix or not. But I think most don't use it, e.g. [i6_core](https://github.com/rwth-i6/i6_core/blob/main/returnn/training.py), which is maybe most relevant.

As the main user so far of this feature (#91), I think this gives some priority to my own preference here, i.e. like I implemented it here in the PR. :)

### DIFF
--- a/sisyphus/loader.py
+++ b/sisyphus/loader.py
@@ -43,6 +43,11 @@ class ConfigManager:
 
         toolkit.set_root_block(filename)
 
+        # maybe remove import path prefix such as "recipe/"
+        for load_path in gs.IMPORT_PATHS:
+            if load_path.endswith('/') and filename.startswith(load_path):
+                filename = filename[len(load_path):]
+                break
         filename = filename.replace(os.path.sep, '.')  # allows to use tab completion for file selection
         assert all(part.isidentifier() for part in filename.split('.')), "Config name is invalid: %s" % filename
         module_name, function_name = filename.rsplit('.', 1)


critias:
So it bothers you that you now have the `recipe/` prefix if use tab completion right?
Couldn't you just add `recipe` to your `gs.IMPORT_PATHS` which would allow sisyphus to find the module given it's path?
Or just move `i6_experiments` into your main directory and add `i6_experiments` to `gs.IMPORT_PATHS`?

albertz:
I just use the default `gs.IMPORT_PATHS`, i.e.:
```python
IMPORT_PATHS = ['config', 'recipe', 'recipe/']
```

Our standard setup (all people at i6) have a `recipe` directory and in there are the repos i6_core, i6_experiments, returnn_common, and maybe others. I would not want to change this structure now.

All the code in i6_core, i6_experiments and so on directly uses `import i6_experiments...`, not with the `recipe.` prefix. We also do not want to change this.

christophmluscher:
I think, the agreed upon convention for `i6_core` is to not add `recipe` to an import. For example: `import i6_core.returnn`. This would then extend to `i6_experiments` as well. 

Just as a suggestion:
Why dont you add a run function in `recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py`. Import the run function into `config.py` and just call `config.run_function`?

albertz:
> Why dont you add a run function in `recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py`. Import the run function into `config.py` and just call `config.run_function`?

Because I want to use the feature from #91, i.e. not using a separate config. Why making it unnecessary complicated?

christophmluscher:
> Why making it unnecessary complicated?

depends on how you set this up if it gets more complicated.. imho. This seems like a workflow preference. Generally I think this should be allowed by sisyphus.

I think the PR otherwise is fine. It extends the functionality of #91 without breaking older workflows

christophmluscher:
I think we should ask: if you have a recipes on top level and in the recipe dir. Should they both be without prefix?! or not?

albertz:
> I think we should ask: if you have a recipes on top level and in the recipe dir. Should they both be without prefix?! or not?

Is this relevant for this PR here? I don't really understand how, and what you mean with your comment. You think there is a case where the PR is bad? What case exactly?

christophmluscher:
> I think the PR otherwise is fine. It extends the functionality of https://github.com/rwth-i6/sisyphus/issues/91 without breaking older workflows

THIS

> Is this relevant for this PR here? I don't really understand how, and what you mean with your comment. You think there is a case where the PR is bad? What case exactly?

no. what I mean is that this could influence some behaviour down the road which we might want to be aware of.

for example
```
sisyphus dir
- recipe
  - some_package
    - write_to_xml function
- different_package_with_same_name
  -write_to_xml function
```

top level imports first?! no imports from top level
maybe this is not really an issue.. just something we might need to keep in mind..

critias:
> I just use the default `gs.IMPORT_PATHS`, i.e.:
> 
> ```python
> IMPORT_PATHS = ['config', 'recipe', 'recipe/']
> ```

Strange, since `recipe` is in your `IMPORT_PATHS` I would expect it to find something like `recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` given that `recipe` is an allowed prefix. If this module then imports stuff without the `recipe` prefix it should also be found since `recipe/` is also part of the `IMPORT_PATHS` am I missing something here?

> Is this relevant for this PR here? I don't really understand how, and what you mean with your comment. You think there is a case where the PR is bad? What case exactly?

It's possible that there is `recipe.recipe.test1` and `recipe.test1` your change would now automatically fall back to `recipe.test1` granted this is an edge case and having a `recipe.recipe` subfolder is not great if `recipe/' and `recipe` are part of `gs.IMPORT_PATHS` but that bug would be really hard to track down.

I simply don't like to mess with the import paths more than necessary and would like to see if there is a cleaner way how to solve it. I already regret many of the heuristics I introduced to this part of the code.

albertz:
> different_package_with_same_name

I would say this is a user error anyway. Anyway, I think for this PR it does not matter, as it is ambiguous and wrong anyway.

albertz:
> > I just use the default `gs.IMPORT_PATHS`, i.e.:
> > ```python
> > IMPORT_PATHS = ['config', 'recipe', 'recipe/']
> > ```
> 
> Strange, since `recipe` is in your `IMPORT_PATHS` I would expect it to find something like `recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` given that `recipe` is an allowed prefix.

It does. That is what I explained. But it uses the wrong `recipe` prefix. We don't want that the module is imported via `recipe` prefix, as also already explained.

albertz:
> It's possible that there is `recipe.recipe.test1` and `recipe.test1` your change would now automatically fall back to `recipe.test1` granted this is an edge case and having a `recipe.recipe` subfolder is not great if `recipe/` and `recipe`are part of`gs.IMPORT_PATHS` but that bug would be really hard to track down.

I really don't think this is any issue, as this will probably break in many other ways as well, as Python/Sisyphus would always get confused where to import things from.

But if you really think this is such an important issue, we can simply introduce a check for this.

Remember that we are discussing here sth which you don't use and would not affect you at all anyway, right?

albertz:
> Since I'm the person most like having to maintain it it does effect me. Messing with the import path might cause unexpected behavior as already explained, at the same time you can solve your tab completion problem with a simple symlink: `ln -s recipe/i6_experiments i6_experiments`

No, because such symlink would cause exactly the issues you explained. I don't want to mess with my setup and introduce such potential problem.

This does not mess with the import path for any of your setups.

Without this PR here, #91 is basically useless now.

albertz:
And I really need it, and other people as well at i6, esp those who would want to use #91.

albertz:
You realize this here is about loading the config?

critias:
> No, because such symlink would cause exactly the issues you explained. I don't want to mess with my setup and introduce such potential problem.

What problem does this symlink cause? The experiment directory is not inside the PYTHONPATH or loaded otherwise.

> Without this PR here, #91 is basically useless now.

Just in case that I misunderstood anything here: Currently you have to type `sis m i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` but you want to be able to type `sis m recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` right?

albertz:
> What problem does this symlink cause? The experiment directory is not inside the PYTHONPATH or loaded otherwise.

As explained above, we use `import i6_experiments...`. This becomes ambiguous now whether it takes the one from the symlink or the one in `recipe`. There should not be such ambiguation.

But also, such workaround really should be unnecessary. I don't really see any reason not to have this PR here. Why make it unnecessary complicated for the user? There is really no downside here. Esp not for anyone not using #91 (so not for you or anyone else currently).

> Just in case that I misunderstood anything here: Currently you have to type `sis m i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` but you want to be able to type `sis m recipe/i6_experiments/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py` right?

Yes, but not only that. Currently the latter also works without throwing an error but potentially this messes up sth, because you probably end up loading many modules twice, once with `recipe` prefix, and then without.

critias:
> > What problem does this symlink cause? The experiment directory is not inside the PYTHONPATH or loaded otherwise.
> 
> As explained above, we use `import i6_experiments...`. This becomes ambiguous now whether it takes the one from the symlink or the one in `recipe`. There should not be such ambiguation.

These is no ambiguity, the symlink is placed inside the work directory which by default is not inside your `PYTHONPATH` and also not in our `IMPORT_PATHS` so it would not be loaded.

> Yes, but not only that. Currently the latter also works without throwing an error but potentially this messes up sth, because you probably end up loading many modules twice, once with `recipe` prefix, and then without.

Calling #91 useless without this commit sound like a exaggeration...
Potentially loading the same module twice is a good point, this could really cause problems! Thinking about this having `recipe` and `recipe/` inside `IMPORT_PATHS` is potentially dangerous independent of this problem, I didn't consider this before 🤔

Does anybody of you really need both in your `IMPORT_PATHS`? Otherwise I would add a warning if `x` and `x/` (probably even `y.startswith('x/')`) are inside `IMPORT_PATHS`. If these collisions are impossible I don't see any potential harm left by this PR.

albertz:
> Thinking about this having `recipe` and `recipe/` inside `IMPORT_PATHS` is potentially dangerous independent of this problem

Is it? It only leads to such issues (modules loaded twice) when the user really mixes both `import recipe.i6_experiments...` and `import i6_experiments`, which you could argue is a user error.

There could also be some check in `RecipeFinder` that there are never such inconsistencies.

> Does anybody of you really need both in your `IMPORT_PATHS`?

I don't. I only need `recipe/`. This is for all people using i6_core and i6_experiments.

But there might be people who use a mix of both. And they might also argue sth like "it worked for me, I did not had any problems", without realizing the potential issues. But I don't know if there is anyone who does use a mix of both. (@michelwi @JackTemaki @ZhouW321 @Marvin84 maybe you know?)

> Otherwise I would add a warning if `x` and `x/` (probably even `y.startswith('x/')`) are inside `IMPORT_PATHS`.

But this is currently the default `IMPORT_PATHS`. So you want to change the default?

critias:
> > Thinking about this having `recipe` and `recipe/` inside `IMPORT_PATHS` is potentially dangerous independent of this problem
> 
> Is it? It only leads to such issues (modules loaded twice) when the user really mixes both `import recipe.i6_experiments...` and `import i6_experiments`, which you could argue is a user error.

I tried a few things, loading the same file using the `recipe/` prefix and not using creates as assumed two different modules, but since the `recipe` is removed in both cases it doesn't effect the hash which was my main concern given that Jobs or other objects inside the config file could . It still isn't pretty, but the only problem that comes to mind is now if some attributes of a module are set for one import it obviously doesn't  effect the other module, e.g.:
```
import recipe.foo as m1
import foo as m2
m1.bar = 1
m2.bar = 2
=>
m1.bar != m2.bar
```

Not great, but I was more worried about the hashes. Using a mix of relative paths and absolut import paths can create a strange mix of which module version is loaded, if this leads to different hashes it could be hard to reproduce...

> There could also be some check in `RecipeFinder` that there are never such inconsistencies.
> 
> > Does anybody of you really need both in your `IMPORT_PATHS`?
> 
> I don't. I only need `recipe/`. This is for all people using i6_core and i6_experiments.
> 
> But there might be people who use a mix of both. And they might also argue sth like "it worked for me, I did not had any problems", without realizing the potential issues. But I don't know if there is anyone who does use a mix of both. (@michelwi @JackTemaki @ZhouW321 @Marvin84 maybe you know?)
> 
> > Otherwise I would add a warning if `x` and `x/` (probably even `y.startswith('x/')`) are inside `IMPORT_PATHS`.
> 
> But this is currently the default `IMPORT_PATHS`. So you want to change the default?

I didn't consider this whole situation when I accepted this change. It would been cleaner to not have this as part of the default Sisyphus version. Everybody could simply put this inside the `settings.py` file if needed.

Anyway, back to the actual PR: After looking at the code I think the only new danger introduced by the PR is that it would change which file is read in a few cases. This does bother me, but it's most likely a rare event and mainly there is the `logging.info('Loaded config: %s' % config_name)` line later on printing exactly which file is used in the end. I think this removes any ambiguity which file is loaded and makes this change not as invisible as I thought.

michelwi:
What about
```
import recipe.foo as m1
import foo as m2
j = m2.SomeJob()
assert isinstance(j, m1.SomeJob)
```
We use this kind of check at some places

albertz:
> What about
> 
> ```
> import recipe.foo as m1
> import foo as m2
> j = m2.SomeJob()
> assert isinstance(j, m1.SomeJob)
> ```
> 
> We use this kind of check at some places

You do? But this check would raise an exception, because `isinstance(j, m1.SomeJob)` is not true here. That's exactly the point, why it's dangerous and probably leads to unexpected behavior to mix both ways. Unless `SomeJob` is not defined in `foo` but comes from somewhere else.

critias:
> ```
> import recipe.foo as m1
> import foo as m2
> j = m2.SomeJob()
> assert isinstance(j, m1.SomeJob)
> ```
> 
> We use this kind of check at some places

I also don't get it. You load the same module with recipe prefix and without recipe prefix on purpose? The given example fails if I test it (as it should).

michelwi:
> I also don't get it. You load the same module with recipe prefix and without recipe prefix on purpose? 

sorry, this was unclear. A less minimal example would be
```
import foo as m2

def test_job(j):
  assert isinstance(j, m2.SomeJob)
```

and now I am afraid that if someone uses their config with recipe prefix, then this would result in something like
```
import recipe.foo as m1
from above_code import test_job

j = m1.SomeJob()
test_job(j)
```

wich fails (as it should technically) but (semantically) the user would expect it to work.

And when I say "I am afraid" that means I did not test if this is in fact a problem in practice and I am not speaking against this change. I was only speaking up because @critias was focusing only on the hashes 

> Not great, but I was more worried about the hashes. Using a mix of relative paths and absolut import paths can create a strange mix of which module version is loaded, if this leads to different hashes it could be hard to reproduce...

and I was just saying that Job instanceship should maybe also be considered.

albertz:
@michelwi if I understand you right, all what you write does not really matter for the PR here, right?

> sorry, this was unclear. A less minimal example would be
> 
> ```
> import foo as m2
> 
> def test_job(j):
>   assert isinstance(j, m2.SomeJob)
> ```

This code is what a user has in his config, or in some recipe? I assume you mean config. And I assume `foo` is part of the recipe.

> and now I am afraid that if someone uses their config with recipe prefix, then this would result in something like
> 
> ```
> import recipe.foo as m1
> from above_code import test_job
> 
> j = m1.SomeJob()
> test_job(j)
> ```
> 
> wich fails (as it should technically) but (semantically) the user would expect it to work.

This would not really have worked anyway. If `foo` is a recipe, e.g. assume `i6_core` instead of `foo` to make it more specific, and `i6_core` was designed in a way that inside `i6_core`, all other modules are imported without `recipe` prefix, which is the case, you cannot do `import recipe.i6_core`. If you would do that, it would not have worked, and you would have some of these effects:

- hashes are broken
- strange mix of some modules loaded multiple times (once with `recipe` prefix, once without)
- strange side effects due to the previous reason

All this was always the case, independent of this PR.

> I was only speaking up because @critias was focusing only on the hashes
> and I was just saying that Job instanceship should maybe also be considered.

Both are kind of the same thing. The hash is directly influenced by the `Job` class name. E.g. `i6_core.returnn.training.ReturnnTrainingJob` has a different hash than `recipe.i6_core.returnn.training.ReturnnTrainingJob`.

PR #106, opened 2022-09-20
typo fix

Atticus1806:
Fixes typo in a word in docstring.


Atticus1806:
While we are at it some more, true -> True is just to keep consistent with "python" naming.

PR #107, opened 2022-09-22
add DelayedJoin

JackTemaki:
Needed for https://github.com/rwth-i6/i6_core/pull/307


critias:
I missed that there where other reviewers connected to this PR as well and just merged it since I didn't see any problems with this PR...
If there are any further comments about this PR just add them here anyway.

Issue #112, opened 2022-10-25
update_engine_rqmt in settings can not be pickled


albertz:
I defined a `update_engine_rqmt` function in my `settings.py`, which is in the root of my setup.
Now I get this pickling error:
```
...
  File "/u/zeyer/setups/combined/2021-05-31/tools/sisyphus/sisyphus/manager.py", line 406, in Manager.run_jobs.<locals>.f
    line: job._sis_setup_directory() 
    locals: 
      job = <local> Job<work/i6_core/returnn/search/ReturnnSearchJobV2.ooVjHv7YzCd9> 
      job._sis_setup_directory = <local> <bound method Job._sis_setup_directory of Job<work/i6_core/returnn/search/ReturnnSearchJobV2.ooVjHv7YzCd9>> 
  File "/u/zeyer/setups/combined/2021-05-31/tools/sisyphus/sisyphus/job.py", line 290, in Job._sis_setup_directory 
    line: pickle.dump(self, f) 
    locals: 
      pickle = <global> <module 'pickle' from '/work/tools/asr/python/3.8.0/lib/python3.8/pickle.py'> 
      pickle.dump = <global> <built-in function dump> 
      self = <local> Job<work/i6_core/returnn/search/ReturnnSearchJobV2.ooVjHv7YzCd9> 
      f = <local> <gzip on 0x7fbcdc4bd190> 
PicklingError: Can't pickle <function update_engine_rqmt at 0x7fbcde8b90d0>: import of module 'settings.py' failed 
```

This is because `update_global_settings_from_file` and `update_global_settings_from_text` do not properly import the module but instead manually do `compile` and `exec`. Esp, the `__name__` for the module exec globals is set incorrectly. Otherwise I think it might have worked already (for the case of `settings.py` in the root dir, at least).

I wonder, is setting `update_engine_rqmt` in the user settings not supported? Or how would that work?

And also, should we maybe fix `update_global_settings_from_text` a bit, that it would at least set `__name__` correctly for this case?

albertz:
> Otherwise I think it might have worked already (for the case of `settings.py` in the root dir, at least).

Ok, this is not enough. It is further a problem that there is no module in `sys.modules` with this name. You could also somehow hack around that. E.g. this code inside my `settings.py` makes it work:
```python

import os
import sys

_root_dir = os.path.dirname(os.path.abspath(__file__))

# https://github.com/rwth-i6/sisyphus/issues/112
__name__ = "settings"
[...]

PR #113, opened 2022-10-27
Loading settings file using importlib

critias:
This PR should fix the bug report in issue https://github.com/rwth-i6/sisyphus/issues/112

Not well tested yet. Main drawback is that DEFAULT_ENVIRONMENT_KEEP and DEFAULT_ENVIRONMENT_SET can not be accessed anymore inside the settings file. I'm not sure if this is really used by anybody.

This PR also allows having multiple settings files.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -5,11 +5,12 @@ These settings can be overwritten via a ``settings.py`` file in the current dire
 
 # Author: Jan-Thorsten Peter <peter@cs.rwth-aachen.de>
 
-import sys
+import importlib.util
 import logging
+import sys
+
 import sisyphus.hash
 from sisyphus.global_constants import *
-import os
 
 
 def engine():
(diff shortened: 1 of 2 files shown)


michelwi:
> Main drawback is that DEFAULT_ENVIRONMENT_KEEP and DEFAULT_ENVIRONMENT_SET can not be accessed anymore inside the settings file. I'm not sure if this is really used by anybody.

yes, we make heavy use of this.

albertz:
Actually, I wonder now, when not loading the `settings.py` as a module and not overwriting `__name__` nor `__package__` for it, then the issue #112 is probably also solved.

critias:
Looks like just executing the code and passing the globals also solves the problem and needs less code.
There are less checks, but I don't think this should be a problem. Any thoughts?

tests/settings_loader_unittest.py
```diff
@@ -0,0 +1,52 @@
...
+file_test1 = 'foo'
+file_test2 = 4
+def test_function():
+    return 42
+            """)
+
+        gs.update_global_settings_from_file(self.TEST_SETTINGS_FILE)
+        assert gs.file_test1 == 'foo'
+        assert gs.file_test2 == 4
+        assert gs.test_function() == 42
```
michelwi, line 33: Would
```suggestion
        with open(self.TEST_SETTINGS_FILE, 'wt') as f:
            f.write("""
file_test1 = 'foo'
file_test2 = 4
DEFAULT_ENVIRONMENT_SET['PATH'] += ':bar'
def test_function():
    return 42
            """)

        gs.update_global_settings_from_file(self.TEST_SETTINGS_FILE)
        assert gs.file_test1 == 'foo'
        assert gs.file_test2 == 4
        assert gs.DEFAULT_ENVIRONMENT_SET['PATH'][-4:] == ':bar'
        assert gs.test_function() == 42
```
now still work?

critias: Yes it still works, it's now part of the unittests.

sisyphus/global_settings.py
```diff
@@ -317,69 +318,46 @@ def file_caching(path):
...
-    globals()['GLOBAL_SETTINGS_FILE'] = filename
+    global GLOBAL_SETTINGS_FILE_CONTENT
 
-    content = ''
     try:
         with open(filename, encoding='utf-8') as f:
             content = f.read()
     except IOError as e:
+        import logging
+        logging.warning(f"Could not load settings from {filename}")
```
albertz, line 349: Also print `e`? I would like to see the actual exception.

sisyphus/global_settings.py
```diff
@@ -317,69 +318,46 @@ def file_caching(path):
...
-        if k not in globals_keys:
-            globals()[k] = v
+ENVIRONMENT_SETTINGS = {}
+ENVIRONMENT_SETTINGS_PREFIX = 'SIS_'
 
-    if AUTO_SET_JOB_INIT_ATTRIBUTES:
-        import logging
-        logging.warning('AUTO_SET_JOB_INIT_ATTRIBUTES is deprecated, please set the attributes manually '
-                        'you might want to use self.set_attrs(locals())')
+GLOBAL_SETTINGS_FILE_CONTENT = ''
```
albertz, line 333: Is this purely for the test case? Or would this have any other use?

If purely for the test case, I would maybe make it private (add `_` prefix, and not uppercase).

Also, in any case, add a small comment.

The logic below is also slightly wrong for it when there are multiple setting files. Only the last content would be in there.

albertz (approved):
Despite my comments, looks ok.

sisyphus/global_settings.py
michelwi: > The logic below is also slightly wrong for it when there are multiple setting files. Only the last content would be in there.

As `GLOBAL_SETTINGS_FILE_CONTENT = f"{GLOBAL_SETTINGS_FILE_CONTENT}` + "new settings file content", all files are present.

otherwise agree to your comment,

michelwi (approved):

sisyphus/global_settings.py
albertz: Ah ok I missed that.

critias: It allows to log the exact settings which where used. I use to store them when creating my final report. I'm not doing it right now, but I still think it could be useful.

I changed the code to make it a little clearer that the new content is added to the original string.

sisyphus/global_settings.py
critias: This is only relevant for non existing files, I changed the code and log message to make it more clear.

sisyphus/global_settings.py
albertz: Ok for me then.

sisyphus/global_settings.py
albertz: `IOError` could also be other things, right? But even if normally it's because not found, I think it would just help to print it.

sisyphus/global_settings.py
```diff
@@ -317,72 +318,53 @@ def file_caching(path):
...
-    globals()['GLOBAL_SETTINGS_FILE'] = filename
+    global GLOBAL_SETTINGS_FILE_CONTENT
 
-    content = ''
     try:
         with open(filename, encoding='utf-8') as f:
             content = f.read()
     except IOError as e:
-        if e.errno != 2:
+        if e.errno == 2:
```
albertz, line 349: You could either just catch `FileNotFoundError`, or better use the `errno` module and check for `errno.ENOENT`.

PR #114, opened 2022-11-02
multi_node_slots option for rqmt

albertz:
Maps to parallel environment (PE) for SGE, and to ntasks for Slurm, and other engines will throw an exception.

Needed for https://github.com/rwth-i6/i6_core/pull/333.

### DIFF
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -61,7 +61,9 @@ class SonOfGridEngine(EngineBase):
         :param list[str] ignore_jobs: list of job ids that will be ignored during status updates.
                                       Useful if a job is stuck inside of SGE and can not be deleted.
                                       Job should be listed as "job_number.task_id" e.g.: ['123.1', '123.2', '125.1']
-        :param str pe_name: used to select parallel environment (PE), when parallel_tasks is set in rqmt
+        :param str pe_name: used to select parallel environment (PE), when parallel_tasks is set in rqmt,
+                            as `-pe <pe_name> <parallel_tasks>`.
+                            The default "mpi" is somewhat arbitrarily choosen as we have it in our environment.
         """
         self._task_info_cache_last_update = 0
         self.gateway = gateway
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -61,9 +61,9 @@ class SonOfGridEngine(EngineBase):
         :param list[str] ignore_jobs: list of job ids that will be ignored during status updates.
                                       Useful if a job is stuck inside of SGE and can not be deleted.
                                       Job should be listed as "job_number.task_id" e.g.: ['123.1', '123.2', '125.1']
-        :param str pe_name: used to select parallel environment (PE), when parallel_tasks is set in rqmt,
-                            as `-pe <pe_name> <parallel_tasks>`.
-                            The default "mpi" is somewhat arbitrarily choosen as we have it in our environment.
+        :param str pe_name: used to select parallel environment (PE), when multi_node_slots is set in rqmt,
+                            as `-pe <pe_name> <multi_node_slots>`.
+                            The default "mpi" is somewhat arbitrarily chosen as we have it in our environment.
         """
         self._task_info_cache_last_update = 0
         self.gateway = gateway
@@ -176,8 +176,8 @@ class SonOfGridEngine(EngineBase):
         out.append('-l')
         out.append('h_rt=%s' % task_time)
 
-        if rqmt.get('parallel_tasks', None):
-            out.extend(['-pe', self.pe_name, str(rqmt['parallel_tasks'])])
+        if rqmt.get('multi_node_slots', None):
+            out.extend(['-pe', self.pe_name, str(rqmt['multi_node_slots'])])
 
         qsub_args = rqmt.get('qsub_args', [])
         if isinstance(qsub_args, str):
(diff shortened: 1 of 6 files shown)


sisyphus/son_of_grid_engine.py
```diff
@@ -52,7 +52,7 @@ def try_to_multiply(y, x, backup_value=None):
 
 class SonOfGridEngine(EngineBase):
 
-    def __init__(self, default_rqmt, gateway=None, auto_clean_eqw=True, ignore_jobs=None):
+    def __init__(self, default_rqmt, gateway=None, auto_clean_eqw=True, ignore_jobs=None, pe_name="mpi"):
```
albertz, line 55: Note: This default is somewhat i6 specific, because we configured the PE name to this. However, as i6 is the only real user of SGE, I would argue this is fine?

critias:
So `parallel_tasks` is referring to processes that are communicating using MPI right? If so could we make the name a little bit more specific e.g. `mpi_tasks`?
My first intuition after reading the name was that you try to limit how many tasks are run in parallel.

albertz:
> So `parallel_tasks` is referring to processes that are communicating using MPI right? If so could we make the name a little bit more specific e.g. `mpi_tasks`? My first intuition after reading the name was that you try to limit how many tasks are run in parallel.

It doesn't need to be MPI. MPI is just one tool to make use of the parallel tasks, but there are other things, e.g. Gloo, and using `-pe` in SGE, or `--ntask` in Slurm is not specific to MPI, it works with any other thing as well, like Gloo.

JackTemaki:
I am not sure if `parallel_tasks` is the best name here, as this is in conflict with `concurrent` tasks, meaning that we actually call the Sisyphus worker multiple times on different nodes... for this the `Task` itself also already has the keyword `parallel`. I think this can be very confusing, that `parallel_tasks` in `rqmt` of a Task is something completely unrelated to the `parallel` parameter of a Task.

critias:
> But anyway, I already asked a few times, what name do you suggest then?

Where? I don't see you asking it even once?

But anyway, naming this parameter really isn't trivial given that Task has already a meaning inside of Sisyphus and parallel is already a parameter of Task... My earlier proposed `mpi_task` was obviously a bad idea since we are not only talking about MPI here. How about something like `instances_per_worker`?

albertz:
> > But anyway, I already asked a few times, what name do you suggest then?
> 
> Where? I don't see you asking it even once?

For example here: https://github.com/rwth-i6/i6_core/pull/333#discussion_r1011499478

> `instances_per_worker`

My line of thought was: `rqmt`, and this setting specifically is for the engine, so it should use the common terminology of the engine. Slurm, SGE and others might not totally share the same terminology. But they also agree on many parts. E.g. "job", or "node" means the same thing. "Task" is a bit different though, although both use this term, and a SGE task, Slurm task or Sis task are three different things.

Btw, this table is somehow helpful: https://slurm.schedmd.com/rosetta.pdf Although it's not totally correct. E.g. CPU count is not `-n` in Slurm.

Neither SGE nor Slurm use the terms "instance" nor "worker". So I'm not sure your suggestion is good here. I guess because you want to make it more Sisyphus specific?

Maybe `parallel_workers`? It's somehow a mix between Sis and SGE.

But I also don't really have a very strong opinion on this. So in the end, whatever you prefer.

michelwi:
how about the word "slots"? e.g. `worker_slots`

I would have suggested the word "node" but then multiple workers/instances/slots/execution units might end up on the same node..

sisyphus/son_of_grid_engine.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, default_rqmt, gateway=None, auto_clean_eqw=True, ignore_jobs=
         :param list[str] ignore_jobs: list of job ids that will be ignored during status updates.
                                       Useful if a job is stuck inside of SGE and can not be deleted.
                                       Job should be listed as "job_number.task_id" e.g.: ['123.1', '123.2', '125.1']
+        :param str pe_name: used for parallel environment (PE), when parallel_tasks is set to rqmt
```
michelwi, line 64:
```suggestion
        :param str pe_name: used to select parallel environment (PE), when parallel_tasks is set in rqmt
```
suggest improved wording

sisyphus/son_of_grid_engine.py
michelwi: agreed, this is a reasonable default. All other users can still specify their own name if need be.

michelwi (commented):
apart from the discussion about the naming I am fine with the implementation and would like to give one suggestion for improved wording in docstring.

albertz:
> how about the word "slots"? e.g. `worker_slots`

Ah yes, in one version of the SGE doc, I also found this:

> `-pe parallel-environment number-of-slots`

And in some MPI doc, I also found the term "slot" being used, e.g. here: https://www.open-mpi.org/doc/v4.0/man1/mpirun.1.php#sect3

So using "slots" would also make sense.

> I would have suggested the word "node" but then multiple workers/instances/slots/execution units might end up on the same node..

Yea, "node" would be misleading.

albertz:
Another suggestion: What about `multi_node_slots` or `multi_node_num_slots`? In the literature and in other frameworks, it's often called multi-node multi-GPU training. This name emphasizes that it's exactly about having multiple nodes.

critias:
I think `multi_node_slots` would be a good name, it is not easily confused with other names used here and easy to remember once you know what it does.

The code looks fine to me, but it would be good add this new option for `rqmt` to the `Task` docstring.

albertz:
`multi_node_slots` or `multi_node_num_slots`?

critias:
I think `multi_node_slots` is specific enough (I hope I never regret that statement 😉).

albertz:
Ok, I renamed it. And I also added some short documentation.

Then it is ready?

PR #116, opened 2023-01-26
Fixes CI by fixing the ubuntu version to 20.04

dthulke:
The CI runner currently fails with 
```
Error: Version 3.6 with arch x64 not found
```

ubuntu-latest now defaults to ubuntu 22.04 for which python 3.6 is not available anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


michelwi:
From the README:

> Sisyphus requires a Python >=3.6 installation

Maybe select the lowest available python version to ensure most compatibility?

Or we could change away from ubuntu-latest.

dthulke:
Good point. The alternative is to fix the ubuntu version to 20.04 as we do in returnn.

PR #119, opened 2023-02-23
Changed requirements in the recipes will overwrite automatic requirement updates

critias:
- Some cleanup of the automatic requirement update 
- Changed requirements in the recipes will now overwrite automatic requirement updates.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -62,8 +62,8 @@ def worker_wrapper(job, task_name, call):
 def update_engine_rqmt(last_rqmt: Dict, last_usage: Dict):
     """ Update requirements after a job got interrupted, double limits if needed
 
-    :param dict[str] last_rqmt: requirements that are requested first
-    :param dict[str] last_usage: information about the last usage by the task
+    :param dict[str] last_rqmt: requirements that where requested for previous run of this task
+    :param dict[str] last_usage: information about the used resources of previous run (mainly memory and time)
     :return: updated requirements
     :rtype: dict[str]
     """
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -7,6 +7,7 @@ These settings can be overwritten via a ``settings.py`` file in the current dire
 
 import logging
 import sys
+from typing import Dict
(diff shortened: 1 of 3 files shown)


michelwi:
> * Changed requirements in the recipes will now overwrite automatic requirement updates.

Does this mean once I specify a custom rqmt in my recipe, the actual rqmt isn't doubled any more in case of oom/oot?

Or does it mean if e.g. (i first define 1G of mem rqmt, my job crashes with oom, then i update the recipe to 8G) then the job is resubmitted with 8G rather than 2x 1G?

sisyphus/global_settings.py
```diff
@@ -58,46 +59,33 @@ def worker_wrapper(job, task_name, call):
     return call
 
 
-def update_engine_rqmt(initial_rqmt, last_usage):
-    """ Update requirements after a job got interrupted.
+def update_engine_rqmt(last_rqmt: Dict, last_usage: Dict):
+    """ Update requirements after a job got interrupted, double limits if needed
 
-    :param dict[str] initial_rqmt: requirements that are requested first
+    :param dict[str] last_rqmt: requirements that are requested first
```
patrick-wilken, line 65: Doc string should be something like: requirements requested for the previous run (possibly too little time or memory)

patrick-wilken:
> Or does it mean if e.g. (i first define 1G of mem rqmt, my job crashes with oom, then i update the recipe to 8G) then the job is resubmitted with 8G rather than 2x 1G?

Yes, this. Before, if the submit_log file still existed the last entry would be read and would overwrite the requirements set in the recipes, even if you intentionally updated them yourself to avoid OOM. Now we check whether these user defined requirements changed and if yes always use them. Only in the other case this automatic doubling of requirements is enabled (possibly multiple times if it keeps failing).

Logic looks good to me. I guess we can try it for a few days and then merge?

patrick-wilken (approved):

patrick-wilken:
I realized only now that updating the requirements is also needed when excluding faulty cluster nodes. That finally worked for me right away as expected. And had no other problems using this branch.

PR #121, opened 2023-02-27
Logging more data into "info"

AbdallahNasir:
Stacktrace and aliases are to be printed into the job info file.

These changes were tested on:
1. Multiple values of JOB_ADD_STACKTRACE_WITH_DEPTH.
2. Single config and single aliases for a single job.
3. Single config and multiple aliases for a single job.
4. Multiple configs and multiple aliases for a single job.

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -299,6 +299,9 @@ class Job(metaclass=JobSingleton):
                     f.write('PARAMETER: %s: %s\n' % (key, value))
                 except UnicodeEncodeError as e:
                     f.write('PARAMETER: %s: <UnicodeEncodeError: %s>\n' % (key, e))
+            if self._sis_stacktrace:
+                f.write("STACKTRACE:\n")
+                f.writelines(traceback.format_list(self._sis_stacktrace[0]))
         self._sis_setup_since_restart = True
 
     def __getstate__(self):
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -302,9 +302,9 @@ class Job(metaclass=JobSingleton):
             if self._sis_aliases:
                 for alias in self._sis_aliases:
                     f.write('ALIAS: %s\n' % alias)
-            if self._sis_stacktrace:
+            for stacktrace in self._sis_stacktrace:
                 f.write('STACKTRACE:\n')
-                f.writelines(traceback.format_list(self._sis_stacktrace[0]))
+                f.writelines(traceback.format_list(stacktrace))
         self._sis_setup_since_restart = True
 
     def __getstate__(self):
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -299,8 +299,11 @@ class Job(metaclass=JobSingleton):
                     f.write('PARAMETER: %s: %s\n' % (key, value))
                 except UnicodeEncodeError as e:
                     f.write('PARAMETER: %s: <UnicodeEncodeError: %s>\n' % (key, e))
+            if self._sis_aliases:
+                for alias in self._sis_aliases:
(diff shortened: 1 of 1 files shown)


curufinwe (approved):

curufinwe:
Keeps old info, this file is written when the job directory is setup for the first time

michelwi:
> What happens if the alias is changed after initial creation? Is info updated then or does it keep the old info?

As this function is only called when the job is setup, the info file will only be updated once you clean the folder (e.g. after error, interrrupted_nonresumable, manual delete)

Atticus1806:
Maybe we should add a timestamp in the info to values which might have changed by then.

critias:
> Maybe we should add a timestamp in the info to values which might have changed by then, so the user does not have to look at when the file was created

I don't mind having a timestamp there as well, but it's nothing that I've ever missed... Feel free to create a PR.

The alias and therefore the info file can also depend on which config is loaded etc. That would actually be more important than the creation date.
Anyway, the idea behind adding aliases was just to have an additional hint in your info file. Reading the comments now makes we wonder if it's actually a help or confusing 🤔

Atticus1806:
So I think if the alias is "correct" during creating it will help.
Maybe I am too clumsy, but for me it happens from time to time that I create a new experiment and copy over code from an existing experiment, missing to change some aliases or setting them slightly wrong. Then it might be confusing I think.

AbdallahNasir:
I would need the alias or the stacktrace to track and understand what is going wrong in the code, maybe if the job resulted in an error. If everything went well, I guess I am no longer interested in debugging what was happening. 

Updating the info for finished jobs would be expensive, especially if they were cleaned up.

Atticus1806:
I agree, updating after creating is not feasible. 
In genral I am not oppsing this PR, I was just wondering if there is a way to work arround this. But maybe there is no good way.

critias:
> So I think if the alias is "correct" during creating it will help. Maybe I am too clumsy, but for me it happens from time to time that I create a new experiment and copy over code from an existing experiment, missing to change some aliases or setting them slightly wrong. Then it might be confusing I think.

Well, if you use create wrong alias there is nothing Sisyphus can do 😄 

> Updating the info for finished jobs would be expensive, especially if they were cleaned up.

It shouldn't be done automatically and every time, but it might be a good idea to be able to trigger an info refresh manually. Right now the traceback would only be inside the info file if tracebacks were enabled when the info file was written. Given that tracebacks slow down the start up quite a bit, I expect most people (including me) to have it disabled by default. So if you want to debug something with tracebacks you'll need a way to update the info files or you would not find any traceback inside the info file 🤔
Cleaned up jobs are not a problem here, since the info file will not be removed.

Atticus1806:
>Well, if you use create wrong alias there is nothing Sisyphus can do 😄

That is true :D I was referring to the case: Start Sisyphus, realize its wrong and fix it, still the wrong alias in the file. A manual retrigger would already solve this.

AbdallahNasir:
I think there is another case that might be interesting. I just tested it and I am not sure how to solve it. 
Using async config while having the same inputs for a job that is defined twice! Well, Sisyphus will execute it once, but it was defined twice in the config. If the config does not contain `await` then all job stacktrace and aliases will be there in the `info`. But if we have `await`, then the job was created again, then in the `info` we will not have the new instance trace or aliases. 

```python
    j = DummyJob(2)
    j.add_alias("some_alias")
    await tk.async_run(j)
    j2 = DummyJob(2)
    j2.add_alias("some_alias2") # Does not show in the info
```
@Atticus1806  I guess this too has no good solution 😅 

I would accept the hint in the `info` regardless, if the job has something wrong, it will fail the first time it was called and will not reach the second time after the `await`.

PR #125, opened 2023-03-02
Change default to raise VariableNotSet and only raise it if no backup value is set

critias:


JackTemaki (approved):
The default behavior is now much stricter, but I think this is okay. Maybe one can add a hint on how to disable it but otherwise I think this is a good change. I was not really aware that the backup flag exists, this seems to be a good option for unset variables.

critias:
> The default behavior is now much stricter, but I think this is okay. Maybe one can add a hint on how to disable it but otherwise I think this is a good change. I was not really aware that the backup flag exists, this seems to be a good option for unset variables.

Adding a hint is a good idea, I just did that with the new commit.
We mainly use the backup value to add Variables inside a report even if they are not computed yet.

PR #126, opened 2023-03-02
Replaced example folder with a simpler workflow

critias:
I think the current workflow is very confusing and a bad example to get started with Sisyphus. Here is a much simpler workflow, it doesn't cover as many edge cases, but should be much easier to understand.


example/recipe/tools.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def run(self):
+        line = self.sh('wc < {text}', capture_output=True)
+        l, w, c = line.split()
+        self.lines.set(int(l))
+        self.words.set(int(w))
+        self.character.set(int(c))
+
+    # Here is an example of task returning a generator
+    def tasks(self):
```
michelwi, line 96:
```suggestion
    def tasks(self) -> Iterable[Task]:
```

example/recipe/splitter/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        assert isinstance(text, tk.Path)
+
+        self.text = text
+        self.splitter = splitter
+        self.out_prefix = 'splitted.'
+        self.splitted_dir = self.output_path('out', True)
+
+    # It's unclear how many outputs will be created by this job
+    # A way around it is to compute all output paths after this job has finished
+    async def outputs(self):
```
michelwi, line 20:
```suggestion
    async def outputs(self) -> List[tk.Path]:
```

example/recipe/splitter/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        out = []
+        for path in sorted(os.listdir(str(self.splitted_dir))):
+            if path.startswith(self.out_prefix):
+                out.append(self.output_path('out/' + path))
+        return out
+
+    def run(self):
+        self.sh('cat {text} | {splitter} {splitted_dir}/{out_prefix}')
+
+    def tasks(self):
```
michelwi, line 31:
```suggestion
    def tasks(self) -> Iterable[Task]:
```

michelwi (commented):
It started off with some minor beauty comments but then I wen full typing-nerd.. sorry xD

Overall very instructive example and I have even learned a thing or two.. Was fully bamboozled by the variable number of outputs of the splitter job xD

example/recipe/tools.py
```diff
@@ -0,0 +1,97 @@
+from sisyphus import *
+import os
+from typing import List
+
+
+class Pipeline(Job):
+    """ This job takes a text file as input and pipes it through all given commands """
+
+    # Everything given to the constructor will be used to compute a hash for this job.
+    def __init__(self, text: tk.Path, pipeline: List):
```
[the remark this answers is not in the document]
critias: I thought about it, but it doesn't only allow strings here. It basically allows anything and converts it into strings. The main usage is obviously to use Path, Variables, and DelayedOps here.
I could change it to `List[Any]`, but I thought just leaving it at list would be the same anyway.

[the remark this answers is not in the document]
michelwi: Oh right.. I never thought of having a pipeline with variable pipeline elements xD

> I could change it to List[Any], but I thought just leaving it at list would be the same anyway.

Ok, lets keep it.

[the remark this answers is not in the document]
critias: I added the `Any`, that should make it clear

example/config/workflow.py
```diff
@@ -0,0 +1,49 @@
...
+    # workflow will be stopped at this point and only continue once the paragraphs_job is finished
+    paragraphs = await paragraphs_job.outputs()
+    wc_per_paragraph = [tools.WordCount(paragraph) for paragraph in paragraphs]
+
+    for i, wc in enumerate(wc_per_paragraph):
+        # You can also register variables as output:
+        tk.register_output('paragraph.%02i.words' % i, wc.words)
+
+    # All jobs inside the wc_per_paragraph list will be computed after this line
+    await tk.async_run(wc_per_paragraph)
```
JackTemaki, line 45: Not sure if this can be part of the simple example, but what do you do if you have multiple experiments and you do not want to block "later" experiements that are independent of this run?

critias:
What's missing here is a simple example of having multiple tasks for one Job and a Tasks with multiple parallel threads. Any suggestions? 🤔

PR #127, opened 2023-03-17
path warning only for alias/output/work

JackTemaki:
Fix for https://github.com/rwth-i6/i6_core/issues/364

### DIFF
--- a/sisyphus/job_path.py
+++ b/sisyphus/job_path.py
@@ -65,9 +65,9 @@ class AbstractPath(DelayedBase):
 
         if gs.WARNING_ABSPATH and not hash_overwrite and path.startswith(gs.BASE_DIR):
             if (
-                    path.startswith(os.path.join(gs.BASE_DIR, gs.ALIAS_DIR)) or
-                    path.startswith(os.path.join(gs.BASE_DIR, gs.OUTPUT_DIR)) or
-                    path.startswith(os.path.join(gs.BASE_DIR, gs.WORK_DIR))
+                path.startswith(os.path.join(gs.BASE_DIR, gs.ALIAS_DIR))
+                or path.startswith(os.path.join(gs.BASE_DIR, gs.OUTPUT_DIR))
+                or path.startswith(os.path.join(gs.BASE_DIR, gs.WORK_DIR))
             ):
                 logging.warning('Creating absolute path inside current work directory: %s '
                                 '(disable with WARNING_ABSPATH=False)' % path)
--- a/sisyphus/job_path.py
+++ b/sisyphus/job_path.py
@@ -63,9 +63,14 @@ class AbstractPath(DelayedBase):
                                         Gets path as input and must be pickleable
         """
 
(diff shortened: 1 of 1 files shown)


critias:
I think skipping all these directories is a bad idea. For the given Problem in the mentioned Issue:

Why don't you change `get_executable_path` to only pass a relative path if you have to create new Path? Sisyphus should take care that the relative path is adjusted when it's called inside the Job.

michelwi:
> I think skipping the check for all these directories is a bad idea.

Could you please elaborate on that? I thought that the check is to prevent accidentally re-defining the output of a Job (which would be a relative path with a creator) as an external input to the graph (via an absolute path).
Therefore having absolute paths in places that cannot possibly be the output of a job would be fine.

Also, which directories do you mean with "all these directories"? 
apart from `alias`, `output`, `work`, and `recipe` I do not have any more folders in my typical setup

> For the given problem in the mentioned Issue:
> Why don't you change `get_executable_path` to only pass a relative path if you have to create new Path? Sisyphus should take care that the relative path is adjusted when it's called inside the Job.

Sorry, I do not fully understand your solution, so let me explain the problem in more detail. The `get_executable_path` function already correctly handles inputs that come from other Jobs in the graph as well as external inputs defined by the user. Only when no path was specified by the user, there should be a "sensible fallback". 
For python I selected the python that was used to start Sisyphus as a fallback; more precisely
```
fallback = tk.Path(shutil.which(gs.SIS_COMMAND[0]))
```
using `shutil.which` to handle calls like `python3 sis m ...` without abspath.

In @JackTemaki's case, the problem is that the python he uses to start the manager lives in a virtual env which has the folder _inside_ the setup directory but is otherwise independent of the setup.

Is your suggestion now to use a relative path instead like
```
fallback_str = shutil.which(gs.SIS_COMMAND[0])
if fallback_str.startswith(gs.BASE_DIR):
  fallback_str = f"./{fallback_str[len(gs.BASE_DIR):]}"
fallback = tk.Path(fallback_str)
```
? I can see how this would programmatically avoid triggering the warning, but would this be in the spirit of this check?

critias:
It looks like there is a misunderstanding what this check should warn you against. It's not uncommon to have or link special config files or experiment related data inside the experiment directory. If this input is accidently added as absolute path the hashes will all depend on the location of the experiment folder. This check was added to warn you against that.

So a warning that the absolute path to `venv/bin/python3` was added is exactly what this check should do! And avoiding this check by making the path relative is totally ok.

Removing everything except work, alias, and output would basically make this check useless for what it was meant to do. I guess this warning should contain a little more information 😄

michelwi:
> It's not uncommon to have or link special config files or experiment related data inside the experiment directory.

Oh I see the misunderstanding. Now that you say it I remember that this is the case in the MT recipes, while it is never done in the ASR recipes (exactly because of the hash)

> So a warning that the absolute path to venv/bin/python3 was added is exactly what this check should do! And avoiding this check by making the path relative is totally ok.

I see, thanks for the clarification. I think we will try to solve it then in i6_core.

PR #131, opened 2023-05-10
resolve links during job import

### LINKED ISSUE #130: Symlinks for Imported Jobs
When importing the work dir from someone else, you might end up with links to `/u/<username>/...`. In the case that you work with an apptainer image, this can be problematic because `/u/<username>/` is not mounted. We could resolve this by only linking real paths upon the import of the work directory.

michelwi:
fixes #130

In case the import_path contains a symlink, this will avoid unnecessary hops while copying or symlink directly to the final target.

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -614,6 +614,8 @@ class Job(metaclass=JobSingleton):
                             base_dir = os.path.dirname(local_path)
                             if not os.path.isdir(base_dir):
                                 os.makedirs(base_dir)
+                        # resolve all intermediate links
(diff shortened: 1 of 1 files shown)


vieting (approved):
I just tested it for the case where I had problematic symlinks and the fix works as expected.

I'm not sure if there might be a case where you would actually want to import from a place that is not the real location, but imho we can merge as this will be helpful for all cases where apptainer images are used and home folders are not mounted. We might add a workaround if that is actually needed.

critias (approved):

PR #132, opened 2023-05-11
Add management of local GPUs to localengine

michelwi:
The localengine is missing a way to organize which jobs run on which local GPU.

This adds some minimal management facilities via explicitly setting the CUDA_VISIBLE_DEVICES environment variable for each subprocess requesting GPUs.

### DIFF
--- a/sisyphus/localengine.py
+++ b/sisyphus/localengine.py
@@ -86,6 +86,9 @@ class LocalEngine(threading.Thread, EngineBase):
         self.max_resources = {'cpu': cpus, 'gpu': gpus}
         self.max_resources.update(kwargs)
         self.free_resources = self.max_resources.copy()
+        cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", "")
+        assert gpus == 0 or len(cuda_visible_devices.split(",")) == gpus
+        self.available_gpus = {g: True for g in cuda_visible_devices.split(",") if g != ""}
 
         self.running_subprocess = []
         self.started = False
@@ -109,14 +112,18 @@ class LocalEngine(threading.Thread, EngineBase):
     def get_default_rqmt(self, task):
         return LOCAL_DEFAULTS
 
-    def start_task(self, task):
+    def start_task(self, task, selected_gpus):
         """
         :param TaskQueueInstance task:
+        :param str selected_gpus:
         :rtype: psutil.Process
         """
         # Start new task
         call = task.call[:]
-        sp = subprocess.Popen(call, start_new_session=True)
+        env = os.environ.copy()
+        if len(selected_gpus) > 0:
+            env["CUDA_VISIBLE_DEVICES"] = selected_gpus
+        sp = subprocess.Popen(call, env=env, start_new_session=True)
         self.running_subprocess.append(sp)
         pid = sp.pid
         process = psutil.Process(pid)
@@ -134,7 +141,7 @@ class LocalEngine(threading.Thread, EngineBase):
 
             logging.debug('Check for finished tasks')
             with self.running_tasks as running_tasks:
-                for process, task in list(running_tasks.values()):
+                for process, task, _ in list(running_tasks.values()):
                     logging.debug('Task state: %s %i PID: %s %s' % (task.task_name, task.task_id,
                                                                     process.pid, process.is_running()))
                     if not process.is_running():
@@ -157,12 +164,25 @@ class LocalEngine(threading.Thread, EngineBase):
         for key, max_available in self.max_resources.items():
             free = self.free_resources[key]
             assert 0 <= free <= max_available
+        # reserve specific GPUs
+        selected_devices = []
+        for name, free in self.available_gpus.items():
+            if len(selected_devices) == rqmt.get("gpu", 0):
+                return ",".join(selected_devices)
+            if free:
+                self.available_gpus[name] = False
+                selected_devices.append(name)
+        assert len(selected_devices) == rqmt.get("gpu", 0)
+        return ",".join(selected_devices)
 
-    def release_resources(self, rqmt):
+    def release_resources(self, rqmt, selected_devices):
         self.free_resources = {key: free + rqmt.get(key, 0) for key, free in self.free_resources.items()}
         for key, max_available in self.max_resources.items():
             free = self.free_resources[key]
             assert 0 <= free <= max_available
+        if selected_devices != "":
+            for name in selected_devices.split(","):
+                self.available_gpus[name] = True
 
     @tools.default_handle_exception_interrupt_main_thread
     def run(self):
@@ -184,7 +204,7 @@ class LocalEngine(threading.Thread, EngineBase):
                         with self.running_tasks as running_tasks:
                             # if enough free resources => run job
                             if self.enough_free_resources(next_task.rqmt):
-                                self.reserve_resources(next_task.rqmt)
+                                selected_gpus = self.reserve_resources(next_task.rqmt)
                                 name = (next_task.name, next_task.task_id)
                                 logging.debug('Start task %s' % str(name))
                                 try:
@@ -193,8 +213,8 @@ class LocalEngine(threading.Thread, EngineBase):
                                     logging.warning('Could not delete %s from waiting queue. '
                                                     'This should not happen! Probably a bug...' % str(name))
                                 # Start job:
-                                process = self.start_task(next_task)
-                                running_tasks[name] = (process, next_task)
+                                process = self.start_task(next_task, selected_gpus)
+                                running_tasks[name] = (process, next_task, selected_gpus)
                                 next_task = None
                                 wait = False
 
@@ -232,6 +252,7 @@ class LocalEngine(threading.Thread, EngineBase):
 
     def task_done(self, running_tasks, task):
         name = (task.name, task.task_id)
+        selected_gpus = running_tasks[name][2]
         logging.debug('Task Done %s' % str(name))
         try:
             del running_tasks[name]
@@ -240,7 +261,7 @@ class LocalEngine(threading.Thread, EngineBase):
                 'Could not delete %s from waiting queue. This should not happen! Probably a bug...' % str(name))
 
         # release used resources
-        self.release_resources(task.rqmt)
+        self.release_resources(task.rqmt, selected_gpus)
 
     def task_state(self, task, task_id):
         name = task.task_name()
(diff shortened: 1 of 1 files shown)


sisyphus/localengine.py
```diff
@@ -86,6 +86,9 @@ def __init__(self, cpus=1, gpus=0, **kwargs):
         self.max_resources = {'cpu': cpus, 'gpu': gpus}
         self.max_resources.update(kwargs)
         self.free_resources = self.max_resources.copy()
+        cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", "")
+        assert gpus == 0 or len(cuda_visible_devices.split(",")) == gpus
```
curufinwe, line 90: why the `or` statement?

sisyphus/localengine.py
```diff
@@ -109,14 +112,18 @@ def start_engine(self):
...
         """
         :param TaskQueueInstance task:
+        :param str selected_gpus:
         :rtype: psutil.Process
         """
         # Start new task
         call = task.call[:]
-        sp = subprocess.Popen(call, start_new_session=True)
+        env = os.environ.copy()
+        if len(selected_gpus) > 0:
```
curufinwe, line 124: more pythonic:
```
if selected_gpus:
```
Also: wouldn't it be better to always set it even if no gpu is selected? This way a job cannot without explicitly setting CUDA_VISIBLE_DEVICES itself take the GPU.

sisyphus/localengine.py
michelwi: For zero gpus  `cuda_visible_devices` will be empty string, so `cuda_visible_devices.split(",")` will be `[""]` of length 1

sisyphus/localengine.py
JackTemaki: I think the empty string is the normal way, or not?

michelwi: CUDA_VISIBLE_DEVICES is now set to empty string if no gpu is requested

sisyphus/localengine.py
```diff
@@ -71,7 +71,7 @@ class LocalEngine(threading.Thread, EngineBase):
     CPU and GPU are always checked, all other requirements only if given during initialisation.
     """
 
-    def __init__(self, cpus=1, gpus=0, **kwargs):
+    def __init__(self, cpus=1, gpus=0, available_gpus="", **kwargs):
```
michelwi, line 74: When a worker starts, it apparently (re-)creates its own version of `Localengine`. This leads to a problem with autodetecting the available GPUs from CUDA_VISIBLE_DEVICES 
(because in the worker this is now set to the specific devices that the worker should run on.)

Until anyone has some idea I add this as a new parameter.

sisyphus/localengine.py
```diff
@@ -297,7 +316,7 @@ def try_to_recover_task(self, task, task_id):
 
             with self.running_tasks as running_tasks:
                 name = (task_instance.name, task_id)
-                running_tasks[name] = (process, task_instance)
+                running_tasks[name] = (process, task_instance, "")
```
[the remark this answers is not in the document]
critias: You could try to read it from `/proc/$pid/environ`. I don't know of any nicer interface to that information...
Maybe the current `CUDA_VISIBLE_DEVICES` could be added as part of the usage file. This would be a useful information to have in other cases as well.

sisyphus/localengine.py
```diff
@@ -282,23 +307,19 @@ def try_to_recover_task(self, task, task_id):
             rqmt = d['requested_resources']
             logpath = os.path.relpath(task.path(gs.JOB_LOG_ENGINE))
             call_with_id = task.get_worker_call(task_id)
-            call_with_id += ['--redirect_output']
```
michelwi, line 285: this arg is actually removed in the worker, so no need to add it here
https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/worker.py#L195

sisyphus/localengine.py
```diff
@@ -282,23 +307,19 @@ def try_to_recover_task(self, task, task_id):
...
             name = task.task_name()
             task_name = task.name()
             task_instance = TaskQueueInstance(call_with_id, logpath, rqmt, name, task_name, task_id)
 
-            if call_with_id != process.cmdline()[1:]:
+            if call_with_id[1:] != process.cmdline()[1:]:
```
michelwi, line 314: the first entry in both cases is the python executable, so makes sense to remove it in both cases

sisyphus/localengine.py
```diff
@@ -282,23 +307,19 @@ def try_to_recover_task(self, task, task_id):
...
                 logging.debug('Job changed, ignore this job: %i %s %s' % (pid, process.cmdline(), task_instance.call))
                 return False
 
-            if os.path.abspath(os.getcwd()) != process.cwd():
-                logging.debut('Job changed, ignore this job: %i %s %s' % (pid, os.getcwd(), process.cwd()))
-                return False
-
```
michelwi, line 297: not sure how this ever worked. 
 * `os.path.abspath(os.getcwd())` gives the path where I execute the manager (e.g. `~/setups/mysetup`) 
 * `process.cwd()` is the work folder of a specific job (e.g. `~/setups/mysetup/work/package/MyJob.somehash/work`)

sisyphus/localengine.py
critias: There is a whole block of code around that line which will only be executed if '--redirect_output ' is set, only one step is to remove the argument. Starting here:
https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/worker.py#L185

You're just disabling the redirection of stdout and stderr towards the log file, I doubt that you want to do that.

sisyphus/localengine.py
critias: This line probably worked due to the redirect_output flag. I think a side effect of calling redirect_output is that it creates a subprocess to actually run the worker. Only the subprocess changes it's work directory, so calling cwd on the wrapper process should return the experiment directory.
I have to admit that I didn't use this property intentionally :)

sisyphus/localengine.py
michelwi: I think this is a misunderstanding. Here I remove this in the `try_to_recover_task` function where the `call_with_id` is only created to compare to the original call.

The original call still contains the `--redirect_output` flag. But in the worker, after handling the flag in the code you mentioned, it is removed before the subprocess is spawned. https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/worker.py#L195 Therefore it is not part of the args that the subprocess knows about and must not be added here.

critias: Ah, thought this was used to create the worker job. There should be two worker jobs if redirect_output is enabled, one that only redirects the output to the log file (containing the redirect_output flag) and one that does the actual work (doesn't have the redirect_output). 
I thought we are matching against the one with the redirect_output flag, but I was probably wrong about that. The process_id is read from the process logging file (<job_dir>/usage.*) which should be created by the process that doesn't have the redirect_output flag. On the other hand I'm fairly sure it worked at some point 🤔

michelwi: >  There should be two worker jobs if redirect_output is enabled [...] I thought we are matching against the one with the redirect_output flag

Yes and no. In the localengine, when the task is first executed, the PID of the worker with redirect_output flag is stored. cf
https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/localengine.py#L196-L197
https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/localengine.py#L119-L123

But then in `try_to_recover_task` we read the pid from the PLOGGING file
https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/localengine.py#L276-L278
which will be written from the worker that actually runs the task, i.e. without the redirect_output and with the workdir set to the Jobs workdir.
cf. [`LoggingThread.log_usage`](https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/worker.py#L84) as created in [`worker_helper`](https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/worker.py#L233C5-L233C20) in the part that does not have the redirect_output flag

So currently the PID in `running_tasks[name]` changes when the manager is restarted. (With and without my changes)

> On the other hand I'm fairly sure it worked at some point 🤔

Maybe this was before the redirect_output was introduced in https://github.com/rwth-i6/sisyphus/commit/c4f825b043bfd81c75d40f7fd9719e7ffb375f42? after a brief glance in the state after this commit I still see the same code path as indicated above..


Aside from the historical considerations how the current state came to be, do you agree that the current state is not working as intended (Tasks are never recovered) and that this should be fixed?
And do you find my proposed implementation a suitable fix for the problem? 
Or should we rather try to recover the original PID of the process with redirect_output flag?

sisyphus/localengine.py
critias: Yeah, that was probably wrong...

sisyphus/localengine.py
michelwi: > Maybe it would be good to change the behaviour if a mismatched job is found. Keep the warning that the job changed, but still accept it as the correct job instead of ignoring it.

done now.

sisyphus/localengine.py
```diff
@@ -282,23 +307,18 @@ def try_to_recover_task(self, task, task_id):
...
             logpath = os.path.relpath(task.path(gs.JOB_LOG_ENGINE))
             call_with_id = task.get_worker_call(task_id)
-            call_with_id += ['--redirect_output']
             name = task.task_name()
             task_name = task.name()
             task_instance = TaskQueueInstance(call_with_id, logpath, rqmt, name, task_name, task_id)
 
-            if call_with_id != process.cmdline()[1:]:
+            if call_with_id[1:] != process.cmdline()[1:]:
                 logging.debug('Job changed, ignore this job: %i %s %s' % (pid, process.cmdline(), task_instance.call))
```
critias, line 315: This message is now a little misleading, it should be something like "Job change, recovering it anyway". I think we should also promote this from a debug to a warning message.

critias (approved):
LGTM beside the "Job changed" message.

sisyphus/localengine.py
michelwi: I left the more technical message as debug and added a more sensible warning containing the name of the job in question .

PR #134, opened 2023-05-24
Allow setting short job names

vieting:
The recent change to slurm on our system brought up a painpoint of sisyphus that I had already faced before. We typically have a fixed width for each column when checking the queuing system's status. The job name's column is already the widest of them. However, with my current width of 64 characters  is still not enough, e.g.
```
i6_core.recognition.advanced_tree_search.AdvancedTreeSearchJob.H
```
for a recognition job. I could increase that further, but a nicer alternative could be to change the default name that we give to jobs from the naming scheme above to just
```
AdvancedTreeSearchJob.HckKXQdAFa09
```

In this PR, I propose to introduce `SHORT_JOB_NAMES` as a flag in the global setting which does exactly that.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -18,10 +18,6 @@ ENGINE_NAME = 'slurm'
 TaskInfo = namedtuple('TaskInfo', ["job_id", "task_id", "state"])
 
 
-def escape_name(name):
-    return name.replace('/', '.')
-
-
 def try_to_multiply(y, x, backup_value=None):
     """ Tries to convert y to float multiply it by x and convert it back
     to a rounded string.
@@ -55,7 +51,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
                                       Useful if a job is stuck inside of Slurm and can not be deleted.
                                       Job should be listed as "job_number.task_id" e.g.: ['123.1', '123.2', '125.1']
         :param Callable job_name_mapping: mapping for job names in sbatch
-                                          Example mapping: 'path/to/file/JobName.H4sH' to 'JobName.H4sH'
+                                          Example mapping: 'path/to/file/JobName.H4sH.task' to 'JobName.H4sH.task'
                                           Warning: If the mapping is changed, the engine cannot recognize already
                                           running jobs anymore.
         """
@@ -210,9 +206,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         :param int end_id:
         :param int step_size:
         """
-        if self.job_name_mapping is not None:
-            name = self.job_name_mapping(name)
-        name = escape_name(name)
+        name = self.process_task_name(name)
         sbatch_call = [
             'sbatch',
             '-J',
@@ -313,9 +307,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         """
 
         name = task.task_name()
-        if self.job_name_mapping is not None:
-            name = self.job_name_mapping(name)
-        name = escape_name(name)
+        name = self.process_task_name(name)
         task_name = (name, task_id)
         queue_state = self.queue_state()
         qs = queue_state[task_name]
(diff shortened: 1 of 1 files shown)


critias:
This change would collapse the whole directory hierarchy inside the work folder. I don't this this would be a good idea.
So if I get this right the problem is that the job name is to long for a readable output of the slurm status right?

In that case it might be a better idea to just reduce the length inside the slurm engine 🤔 e.g. similar to the escape_name function:
https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/simple_linux_utility_for_resource_management_engine.py#L20

vieting:
Hmm right, flattening the hierarchy in the work folder is indeed not intended. So you would vote for an engine-specific solution? I pushed it for the slurm engine.

critias:
It would be cleaner to make `escape_name` a method of SimpleLinuxUtilityForResourceManagementEngine instead of passing the `short_job_name` flag.

`escape_name` should be renamed to reflect that it now doesn't only escape the name.

Instead of passing a very specific `short_job_names` parameter you could also just pass a function. That would allow for greater flexibility in your config file, e.g. keeping the last two segments of the path or the first and last segment etc.

I assume that the engine will not be able to match already running jobs after changing the name mapping, so you should add a warning about that (after verifying that this is true).

vieting:
> I assume that the engine will not be able to match already running jobs after changing the name mapping, so you should add a warning about that (after verifying that this is true).

Yes, I also think that already running jobs cannot be matched. How do you envisage this warning? How would we distinguish between running jobs from before changing the mapping function and other running jobs that were submitted independently of this sisyphus setup (e.g. another setup, interactive login sessions, manually submitted jobs etc.)? If this warning is too general, it might be annoying when intentionally using a custom mapping and it still shows up. Or do you mean a warning in the doctring for that option? This could already be sufficient because users actively need to make a change and it would be their responsibility to take care of running jobs in the same setup.

critias:
I was only thinking about warning in the documentation. As you mentioned, everything else would be hard to do.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, default_rqmt, gateway=None, has_memory_resource=True, auto_cl
         :param list[str] ignore_jobs: list of job ids that will be ignored during status updates.
                                       Useful if a job is stuck inside of Slurm and can not be deleted.
                                       Job should be listed as "job_number.task_id" e.g.: ['123.1', '123.2', '125.1']
+        :param job_name_mapping: mapping for job names in sbatch, e.g., path/to/file/JobName.H4sH to JobName.H4sH
```
vieting, line 56: Not sure how to annotate the type here. I'd usually use `typing.Callable`, but it seems like `typing` isn't used in general here.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -204,6 +210,8 @@ def submit_helper(self, call, logpath, rqmt, name, task_name, start_id, end_id,
         :param int end_id:
         :param int step_size:
         """
+        if self.job_name_mapping is not None:
```
critias, line 213: You are now repeating the same three lines twice. It would be better to have method which calls self.job_name_mapping (if it's not None) and does the escaping.

You can delete the escape_name function afterwards since it's not needed anymore.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -2,6 +2,7 @@
 
 from collections import defaultdict, namedtuple
 from enum import Enum
+from typing import Callable
```
critias, line 5: This import isn't needed if Callable is only mentioning in the comments and not used as type hint.

critias (commented):
I don't have any other comments, but it would be nice to have an opinion of some who is actually using slurm as well.
@michelwi @JackTemaki

michelwi:
> I don't have any other comments, but it would be nice to have an opinion of some who is actually using slurm as well. @michelwi @JackTemaki

Looks good to me, but I am currently not using SLURM, so I can't test this.

JackTemaki:
I would use this, will test next week when I start new jobs!

sisyphus/simple_linux_utility_for_resource_management_engine.py
vieting:
```suggestion
```

vieting:
@JackTemaki can you maybe test (and merge)?

JackTemaki:
> @JackTemaki can you maybe test (and merge)?

I tried to, but I had issues with using a local Sisyphus version with the apptainer image (Sisyphus is installed in the image)

michelwi (approved):
> I'm using it without issues. Does anyone else volunteer to test or should we proceed with merging without that?

It ran for weeks without issues, I think we can merge now.

JackTemaki (approved):

PR #137, opened 2023-07-17
sis_hash_helper, do not pickle from __main__

albertz:


sisyphus/hash.py
```diff
@@ -99,8 +99,10 @@ def sis_hash_helper(obj):
         # Not a nice way to check if the given function is a lambda function, but the best I found
         # assert not isinstance(lambda m: m, LambdaType) is true for all functions
         assert obj.__name__ != '<lambda>', "Hashing of lambda functions is not supported"
+        assert obj.__module__ != "__main__", "Hashing of functions defined in __main__ is not supported"
         byte_list.append(sis_hash_helper((obj.__module__, obj.__qualname__)))
     elif isclass(obj):
+        assert obj.__module__ != "__main__", "Hashing of functions defined in __main__ is not supported"
```
albertz, line 105:
```suggestion
        assert obj.__module__ != "__main__", "Hashing of classes defined in __main__ is not supported"
```

albertz:
> Why is it not supported?

Well, maybe "supported" is misleading. I just copied the message from the line before, where it says "Hashing of lambda functions is not supported". It's misleading exactly in the same way. Of course you can give a hash for it and remove that `assert`. However, that hash is probably not what you want because whenever you have a function defined in your `__main__`, it is very likely that this function will not be accessible like this from anywhere else, so in other cases, it would have a different hash, or it would just not be available.

albertz:
> so in other cases, it would have a different hash

To expand on that: This is exactly a case I had, where the hash was suddenly different, because in one case, it had `__main__` as the module (because I imported the module via `python -m ...`) and in the other case it had the full module name.

JackTemaki:
Can you give a link to your setup where you are doing that?

albertz:
`python -m ...` is what I run from command line. You can do that with any file.

albertz:
In this line, you will get a different hash, depending how the code is executed (via `python -m <this_file>` or not but just imported): https://github.com/rwth-i6/i6_experiments/blob/34f329e7219b07fab0e1dc3654978961126a895c/users/zeyer/experiments/exp2023_04_25_rf/_moh_att_2023_06_30_import.py#L58

albertz:
It's just exactly the same reason that we have that `assert` in the line before. To avoid cases the user likely did not expect. I don't think there is any valid reason where you want to import sth from `__main__`?

critias:
LGTM

michelwi (approved):
LGTM.

any further comments @JackTemaki ?

JackTemaki (approved):

PR #139, opened 2023-07-18
respect "required" in OutputReport

michelwi:
The parameter `required` of `OutputReport` was previously unused. 
The same parameter is used by `OutputCall` to explicitly list all inputs that need to be finished in order for the Call to run - e.g. in case the automatic detection fails. I assume the same functionality was planned for `OutputReport` as well.

Point of contention: the function `update_values(self, report_values)` currently does not have such functionality. Should I add it?


critias:
I don't see any reason to not have required as parameter for update_values, but since nobody missed it so far it doesn't look like a high priority :)

PR #140, opened 2023-07-30
Add option to not print Jobs in HOLD state

michelwi:
Sometimes you know that a NN training will not be good, so you cancel it after an early epoch.
One could then remove it from the graph, but 
* this is rather complicated if one uses nested for loops to setup all the jobs
* makes it intransparent why this experiment was removed (i.e. we might still want to see the bad scores of the early epochs)

The next best option is to manually mark the training as HOLD (or ERROR), but then we will still be reminded in the manager about this job, that we do not want any more. (can be really a lot of lines sometimes)

So here I propose a setting `PRINT_HOLD` (default `True` to preserve the old behavior). The manager will skip printing lines for held Jobs if this is set to False. The number of held jobs will still be printed in the overview to remind us of these jobs.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -204,6 +204,8 @@ PRINT_ERROR = True
 PRINT_ERROR_TASKS = 1
 #: Print that many last lines of error state log file
 PRINT_ERROR_LINES = 40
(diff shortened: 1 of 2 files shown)


critias (approved):
LGTM

PR #145, opened 2023-08-09
Check for class instances of AbstractPath in extract_paths

curufinwe:
For some reason I have a setup where extract_paths will return the class tk.Path. Not sure why that is in my setup, but in principle we should check for that and not return the class itself if it happens to be in the object.


michelwi (approved):

critias (approved):
Today I learned that type is not just a function.

PR #162, opened 2023-12-15
Task.run wait for inputs

### LINKED ISSUE #159: `Task.run`, wait for inputs
Some of my recog jobs fail because they don’t find the checkpoint:
```[2023-10-22 14:05:41,761] INFO: Start Job: Job<work/i6_core/returnn/forward/ReturnnForwardJobV2.3UMtLSktznsC> Task: create_files
[2023-10-22 14:05:41,761] INFO: Inputs:
[2023-10-22 14:05:41,761] INFO: /u/zeyer/setups/combined/2021-05-31/tools/returnn
[2023-10-22 14:05:41,761] INFO: /work/tools/users/zeyer/linuxbrew/opt/python@3.11/bin/python3.11
[2023-10-22 14:05:41,761] INFO: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/oggzip/BlissToOggZipJob.NSdIHfk1iw2M/output/out.ogg.zip
[2023-10-22 14:05:41,761] INFO: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/training/ReturnnTrainingJob.jYUaQ42BWrqA/output/models/epoch.240.pt
[2023-10-22 14:05:41,761] WARNING: Input path does not exist: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/training/ReturnnTrainingJob.jYUaQ42BWrqA/output/models/epoch.240.pt
...
[2023-10-22 14:05:41,762] INFO: ------------------------------------------------------------
[2023-10-22 14:05:41,762] INFO: Starting subtask for arg id: 0 args: []
[2023-10-22 14:05:41,762] INFO: ------------------------------------------------------------
[2023-10-22 14:05:41,777] INFO: Run time: 0:00:00 CPU: 215.80% RSS: 70MB VMS: 213MB 
reformatted ...
[2023-10-22 14:05:41,950] ERROR: Job failed, traceback:
...
  File "/u/zeyer/setups/combined/2021-05-31/recipe/i6_core/returnn/forward.py", line 307, in ReturnnForwardJobV2.create_files
    line: assert os.path.exists( 
              _get_model_path(self.model_checkpoint).get_path()
          ), f"Provided model checkpoint does not exists: {self.model_checkpoint}"
...
AssertionError: Provided model checkpoint does not exists: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/training/ReturnnTrainingJob.jYUaQ42BWrqA/output/models/epoch.240.pt
```
then it ends up in this state:
```
interrupted_not_resumable: Job<work/i6_core/returnn/forward/ReturnnForwardJobV2.3UMtLSktznsC>
```

My initial suggestion was to add some wait time in the `ReturnnForwardJobV2` (see https://github.com/rwth-i6/i6_core/issues/464), but @michelwi suggested to implement this more generic, so this issue here is to discuss about this.

> I also found this code:
> https://github.com/rwth-i6/sisyphus/blob/cbf529c36c1f247dc0400d5f5485fde240c6e612/sisyphus/task.py#L140-L146

Maybe we should change actually this code to not just print this warning ("Input path does not exist") but instead to wait in this case?

I really don't like the `WAIT_PERIOD_JOB_FS_SYNC`. I want to get rid of any arbitrary sleeps before doing anything. See also https://github.com/rwth-i6/sisyphus/issues/146. Ideally, if sth is ready, it should directly execute the next task without any delay.

Then, the task itself can check if inputs are available, and if not, wait a bit. Thus, in the ideal case, it would directly run, and only if not available, it would wait a bit.

_Originally posted by @albertz in https://github.com/rwth-i6/i6_core/issues/464#issuecomment-1829627664, after discussion with @michelwi_

albertz:
New options:

- `WAIT_PERIOD_FOR_INPUTS_AVAILABLE`
- `TASK_INPUTS_MUST_BE_AVAILABLE`

Fix #159.

Note that `TASK_INPUTS_MUST_BE_AVAILABLE` is True by default, which can potentially break some old setups, where jobs used an input which was not an existing file but only the prefix of it (e.g. checkpoint file name).

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -195,6 +195,11 @@ WAIT_PERIOD_HTTP_RETRY_BIND = 10
 WAIT_PERIOD_JOB_CLEANUP = 10
 #: How many seconds should all inputs be available before starting a job to avoid file system synchronization problems
 WAIT_PERIOD_MTIME_OF_INPUTS = 60
+#: How long to wait for all inputs to be available in Task.run (https://github.com/rwth-i6/sisyphus/issues/159)
+WAIT_PERIOD_FOR_INPUTS_AVAILABLE = 60
+
+#: Fail when not all inputs are available after WAIT_PERIOD_FOR_INPUTS_AVAILABLE
+TASK_INPUTS_MUST_BE_AVAILABLE = True
 
 #: set true to automatically clean jobs in error state and retry
 CLEAR_ERROR = False
--- a/sisyphus/task.py
+++ b/sisyphus/task.py
@@ -1,3 +1,4 @@
+import math
 import os
 import logging
 import sys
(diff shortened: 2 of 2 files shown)


michelwi:
> Note that TASK_INPUTS_MUST_BE_AVAILABLE is True by default, which can potentially break some old setups, where jobs used an input which was not an existing file but only the prefix of it (e.g. checkpoint file name).

I personally am fine with the new default behavior. But I would like to check with @patrick-wilken if the MT recipes make active use of this "feature".  If yes, then we should rather have FALSE the default.

patrick-wilken:
As said on Slack, I would be okay with having to set `TASK_INPUTS_MUST_BE_AVAILABLE` to make our current recipes work. `WAIT_PERIOD_FOR_INPUTS_AVAILABLE=0` would then also make sense until we stop using path prefixes, right?

michelwi (approved):
> As said on Slack, I would be okay with having to set TASK_INPUTS_MUST_BE_AVAILABLE to make our current recipes work. 

ok. 

> WAIT_PERIOD_FOR_INPUTS_AVAILABLE=0 would then also make sense until we stop using path prefixes, right?

yes, unless you want to wait unnecessarily for each expected-missing input.

PR #181, opened 2024-02-10
Log finished outputs

### LINKED ISSUE #152: Log file for finished output
I often would miss the "Finished output: ..." log message by the Sis manager on stdout, as it usually prints quite a lot when there are a few experiments running.

A log file with those (human readable timestamp + what output was finished) would be very useful.

I assume this would need to be implemented. So this is a feature request. Maybe you could also suggest already a place where and how to implement it, so then I could also do a PR myself.

albertz:
Fix #152.

I used `FINISHED_LOG` as the name for the option to have it consistent with `MEMORY_PROFILE_LOG`.

I used `"log/finished.log"` as the default path to have it consistent with `"log/manager.log.%s"`
from the manager UI.

I thought about whether to have this enabled or disabled by default, and decided that this should not add too much overhead, so we can have this enabled by default.

### DIFF
--- a/sisyphus/graph.py
+++ b/sisyphus/graph.py
@@ -16,6 +16,7 @@ import pprint
 import threading
 from typing import DefaultDict, Optional, List
 from multiprocessing.pool import ThreadPool
+from datetime import datetime
 
 
 class Node(object):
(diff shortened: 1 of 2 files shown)


sisyphus/graph.py
```diff
@@ -112,6 +113,15 @@ def run_when_done(self, write_output=None):
                     # Set new link if needed
                     os.symlink(os.path.realpath(self._sis_path.get_path()), outfile_name)
                     logging.info("Finished output: %s" % outfile_name)
+
+                    if gs.FINISHED_LOG:
+                        if "/" in gs.FINISHED_LOG and not os.path.exists(os.path.dirname(gs.FINISHED_LOG)):
```
critias, line 118: You could also use:
```PY3
if os.path.dirname(gs.FINISHED_LOG):
  os.makedirs(os.path.dirname(gs.FINISHED_LOG)), exist_ok=True)
```

critias (approved):
LGTM

sisyphus/graph.py
albertz: Yes, I know, but my code should be more efficient:
* `dirname` will do some more stuff than just `"/" in gs.FINISHED_LOG`
* `makedirs` is optimized for the case that the dir does not exist yet. However, we will have the case that in almost all cases, the dir does already exist. In this case, `makedirs` has some overhead, and I wanted to avoid that.

PR #182, opened 2024-03-04
add quotation marks to --wrap for sbatch

JackTemaki:
For some reason calling sbatch would interpret the `--wrap` string differently for one of my new setups.

```
sbatch [...] --wrap=apptainer exec --nv --bind /work/asr4 /work/asr4/rossenbach/rescale/pytorch_mixed_precision/apptainer/u22_pytorch2.1_onnx_flashlight_0224_jaist_project.sif sis worker --engine long work/i6_core/returnn/forward/ReturnnForwardJobV2.xT8obDvtAg1w run
[2024-03-04 10:16:31,705] ERROR: Error: sbatch: error: Script arguments not permitted with --wrap option
```

I added quotation marks solving this issue.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -228,7 +228,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
 
         sbatch_call += ["-a", "%i-%i:%i" % (start_id, end_id, step_size)]
         command = '"' + " ".join(call) + '"'
-        sbatch_call += ["--wrap=%s" % " ".join(call)]
+        sbatch_call += ["--wrap='%s'" % " ".join(call)]
         while True:
             try:
                 out, err, retval = self.system_call(sbatch_call)


albertz (approved):
I wonder what aspect is actually interpreting this. Is this some subshell by `sbatch`? Or some custom code by `sbatch`? Or something else?

There is also [`shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote), to escape a string for (POSIX) shells. Using the quoting here could break if there are other quotes in the `call`.

But as we don't really know, and this problem might be unlikely, I think this solution here is good enough.

albertz:
I think we need to revert this. Now all my jobs crash with this error:
```
--------------------- Slurm Task Prolog ------------------------
/var/spool/slurm/slurmd/job5393565/slurm_script: 4:
/work/tools/users/zeyer/py-envs/py3.11-torch2.1/bin/python3.11 ./sis
worker --engine long
work/i6_core/returnn/forward/ReturnnForwardJobV2.CJwh9MSDhDIk run: not
found
--------------------- Slurm Task Epilog ------------------------
```

PR #185, opened 2024-04-19
Improving the scheduling logic of the local engine

dthulke:
In the current implementation the LocalEngine always tries to schedule the first runnable task in its input queue (fifo). In cases where this jobs requires resources that are currently not available (e.g. a GPU), the scheduler waits till the resources are available even if there are smaller jobs that could be scheduled now.

As a very simple fix, with this change, if a task cannot be scheduled it is moved to the end of the queue and the next task is checked.

### DIFF
--- a/sisyphus/localengine.py
+++ b/sisyphus/localengine.py
@@ -203,11 +203,10 @@ class LocalEngine(threading.Thread, EngineBase):
                 logging.debug("Check for new tasks (Free resources %s)" % self.free_resources)
                 # get object for synchronisation
                 with self.waiting_tasks as waiting_tasks, self.runnable_tasks as runnable_tasks:
-                    total_runnable_tasks = len(runnable_tasks)
                     runnable_task_idx = 0
 
                     # run next task if the capacities are available
-                    while runnable_task_idx < total_runnable_tasks:
+                    while runnable_task_idx < len(runnable_tasks):
                         next_task = runnable_tasks[runnable_task_idx]
                         with self.running_tasks as running_tasks:
                             # if enough free resources => run job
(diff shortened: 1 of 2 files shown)


critias:
Good point! It bothers me a little that this now changes the order of the queue if it finds a job which needs less resources. It would be nice to keep the FIFO property.

You could just pull everything out of the Queue and store it in a normal list. Since the LocalEngine has only one thread checking for new tasks it shouldn't be a problem that a normal list isn't thread safe like the Queue. You can than simply iterate over the list and remove submitted jobs from it.

michelwi (approved):

sisyphus/localengine.py
```diff
@@ -194,21 +194,21 @@ def release_resources(self, rqmt, selected_devices):
...
+                # check runnable tasks
+                logging.debug("Check for new tasks (Free resources %s)" % self.free_resources)
+                # get object for synchronisation
+                with self.waiting_tasks as waiting_tasks, self.runnable_tasks as runnable_tasks:
+                    total_runnable_tasks = len(runnable_tasks)
+                    runnable_task_idx = 0
 
                     # run next task if the capacities are available
-                    if next_task is not None:
+                    while runnable_task_idx < total_runnable_tasks:
```
critias, line 210: Why not use `len(runnable_tasks)` directly?

critias (commented):
Thanks for changing the PR. I think it looks good like that, but why do you are you using `total_runnable_tasks` instead of just using `len(runnable_tasks)` directly?

sisyphus/localengine.py
dthulke: Good question :D but now I also don't see a reason to introduce the extra variable.

critias (approved):
Thanks for the changes, I like the this version much better :)

PR #189, opened 2024-05-07
Keep hash consistent for enum in newer Python versions

### LINKED ISSUE #188: `sis_hash_helper` on `enum.Enum`s is Python version dependent
The use of `enum.Enum` currently results in different Sisyphus hashes depending on the Python version…

E.g. from i6_core, we have:
```python
class MergeStrategy(enum.Enum):
    SUBCORPORA = 0
    FLAT = 1
    CONCATENATE = 2
```

`sis_hash_helper(MergeStrategy.FLAT)` gives:
* Python 3.11: `b"(MergeStrategy, (dict, (tuple, (str, '__objclass__'), (EnumType, (tuple, (str, 'i6_core.corpus.transform'), (str, 'MergeStrategy')))), (tuple, (str, '_name_'), (str, 'FLAT')), (tuple, (str, '_sort_order_'), (int, 1)), (tuple, (str, '_value_'), (int, 1))))"`
* Python 3.10: `b"(MergeStrategy, (dict, (tuple, (str, '__objclass__'), (EnumMeta, (tuple, (str, 'i6_core.corpus.transform'), (str, 'MergeStrategy')))), (tuple, (str, '_name_'), (str, 'FLAT')), (tuple, (str, '_value_'), (int, 1))))"`

I think most people still use Python <=3.10? Am i the only one using a more recent Python? I would suggest a fix on Sisyphus side to always keep the old python <=3.10 hash of `enum`, even on newer Python versions. That will require a special check for `enum`, but I think that’s ok.

albertz:
Fix #188.

### DIFF
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -64,7 +64,8 @@ def get_object_state(obj):
 
     if isinstance(obj, enum.Enum):
         assert isinstance(state, dict)
-        state.pop("_sort_order_", None)  # compat with Python <=3.10, https://github.com/rwth-i6/sisyphus/issues/188
+        # In Python >=3.11, keep hash same as in Python <=3.10, https://github.com/rwth-i6/sisyphus/issues/188
+        state.pop("_sort_order_", None)
 
     if args is None:
         return state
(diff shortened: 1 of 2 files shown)


sisyphus/hash.py
```diff
@@ -61,6 +62,10 @@ def get_object_state(obj):
         assert args is not None, "Failed to get object state of: %s" % repr(obj)
         state = None
 
+    if isinstance(obj, enum.Enum):
+        assert isinstance(state, dict)
+        state.pop("_sort_order_", None)  # compat with Python <=3.10, https://github.com/rwth-i6/sisyphus/issues/188
```
critias, line 67: compat <= 3.10 could also mean that this would crash otherwise. 
It would be good to make clear that is done to keep the computed hash equal with python <=3.10.

critias (changes requested):
Beside the comment it looks good to me.

critias (approved):

PR #191, opened 2024-05-31
Add timeout to SSH queue scan command

### LINKED ISSUE #190: Too many open file descriptors
Hi, I was using sisyphus today for a big recipe and I got an error in my worker which claimed `too many open files`:

```
OSError: [Errno 24] Unable to synchronously open file (unable to open file: name = <filename>, errno = 24, error message = 'Too many open files', flags = 0, o_flags = 0)
```

However, not only had the worker crashed, but also the manager crashed with this error. Moreover, the terminal (tmux pane) entered a state where every character I typed was converted into a newline character. As a result, I couldn't type any instruction, so I ended up killing the tmux pane and recreating it.

I got into an investigation of what was happening and I developed a really small test:

```
# test.py

def py():
    pass
```

I found out that running the sisyphus manager on the test (bare; without any `settings.py`) opened ~3k files, from my baseline of 524 opened files to 3254 opened files after running `sis m test.py`, according to `lsof | grep <user> | wc`.

Besides that, I had the issue that every job triggered by the manager was adding exactly 105 opened files to the list of open file descriptors. However, I can't reproduce this starting from scratch, which leads me to think that it might be a problem about how our code base interacts with sisyphus (or just from our code base). I'll keep investigating and keep you tuned.

Is this issue about opening too many files by sisyphus intended because of some sisyphus caching strategy or related work? Was this ever addressed?

If you need more details, I'll be glad to provide them. Thank you in advance.

Icemole:
If the SSH key isn't properly set, sisyphus will schedule stacked queue scan commands which will ask for a password. This will lead to subprocess buildup, which eventually crashes the manager with a `Too many open files` error.

I've also changed the hardcoded 30 second interval for an interval depending on `gs.WAIT_PERIOD_BETWEEN_CHECKS`, which I find makes more sense. <- Edit: not anymore, see below.

Fix #190.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -97,7 +97,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
         p = subprocess.run(
-            system_command, stdin=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
+            system_command, input=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
         )
 
         def fix_output(o):
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -86,7 +86,9 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         """
         if self.gateway:
             escaped_command = [shlex.quote(s) for s in command]  # parameters need to be shell safe when sending via ssh
-            system_command = ["ssh", "-x", self.gateway] + [" ".join(["cd", os.getcwd(), "&&"] + escaped_command)]
+            system_command = ["ssh", "-x", self.gateway, "-o", "ConnectTimeout", gs.WAIT_PERIOD_BETWEEN_CHECKS] + [
+                " ".join(["cd", os.getcwd(), "&&"] + escaped_command)
+            ]
         else:
             # no gateway given, skip ssh local
             system_command = command
@@ -95,7 +97,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        out, err = p.communicate(input=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
 
         def fix_output(o):
             """
@@ -289,7 +291,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
     def queue_state(self):
         """Returns list with all currently running tasks in this queue"""
 
-        if time.time() - self._task_info_cache_last_update < 30:
+        if time.time() - self._task_info_cache_last_update < gs.WAIT_PERIOD_BETWEEN_CHECKS:
             # use cached value
             return self._task_info_cache
 
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -99,8 +99,6 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         p = subprocess.run(
             system_command,
             stdin=send_to_stdin,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
             timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
         )
 
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -100,6 +100,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
             p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
         except subprocess.TimeoutExpired:
             logging.warning("Timeout expired for command: %s" % " ".join(system_command))
+            return [], ["TimeoutExpired"], -1
 
         def fix_output(o):
             """
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -96,11 +96,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(
-            system_command,
-            stdin=send_to_stdin,
-            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
-        )
+        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
 
         def fix_output(o):
             """
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -96,7 +96,9 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
+        p = subprocess.run(
+            system_command, stdin=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
+        )
 
         def fix_output(o):
             """
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -96,9 +96,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(
-            system_command, input=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
-        )
+        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
 
         def fix_output(o):
             """
@@ -292,7 +290,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
     def queue_state(self):
         """Returns list with all currently running tasks in this queue"""
 
-        if time.time() - self._task_info_cache_last_update < gs.WAIT_PERIOD_BETWEEN_CHECKS:
+        if time.time() - self._task_info_cache_last_update < 30:
             # use cached value
             return self._task_info_cache
 
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -86,7 +86,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         """
         if self.gateway:
             escaped_command = [shlex.quote(s) for s in command]  # parameters need to be shell safe when sending via ssh
-            system_command = ["ssh", "-x", self.gateway, "-o", "ConnectTimeout", gs.WAIT_PERIOD_BETWEEN_CHECKS] + [
+            system_command = ["ssh", "-x", self.gateway, "-o", "BatchMode=yes"] + [
                 " ".join(["cd", os.getcwd(), "&&"] + escaped_command)
             ]
         else:
@@ -116,9 +116,9 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
                 assert False
             return o[:-1]
 
-        out = fix_output(out)
-        err = fix_output(err)
-        retval = p.wait(timeout=30)
+        out = fix_output(p.stdout)
+        err = fix_output(p.stderr)
+        retval = p.returncode
 
         # Check for ssh error
         err_ = []
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -96,7 +96,10 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        try:
+            p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        except subprocess.TimeoutExpired:
+            logging.warning("Timeout expired for command: %s" % " ".join(system_command))
 
         def fix_output(o):
             """
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -94,10 +94,15 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
+        p = subprocess.run(
+            system_command,
+            stdin=send_to_stdin,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
+        )
 
         def fix_output(o):
             """
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -97,11 +97,7 @@ class SonOfGridEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(
-            system_command,
-            stdin=send_to_stdin,
-            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
-        )
+        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
 
         def fix_output(o):
             """
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -95,10 +95,15 @@ class SonOfGridEngine(EngineBase):
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
+        p = subprocess.run(
+            system_command,
+            stdin=send_to_stdin,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
+        )
 
         def fix_output(o):
             """
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -100,8 +100,6 @@ class SonOfGridEngine(EngineBase):
         p = subprocess.run(
             system_command,
             stdin=send_to_stdin,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
             timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
         )
 
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -97,7 +97,10 @@ class SonOfGridEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        try:
+            p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        except subprocess.TimeoutExpired:
+            logging.warning("Timeout expired for command: %s" % " ".join(system_command))
 
         def fix_output(o):
             """
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -97,7 +97,9 @@ class SonOfGridEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
+        p = subprocess.run(
+            system_command, stdin=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
+        )
 
         def fix_output(o):
             """
--- a/sisyphus/son_of_grid_engine.py
+++ b/sisyphus/son_of_grid_engine.py
@@ -97,9 +97,7 @@ class SonOfGridEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        p = subprocess.run(
-            system_command, input=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS
-        )
+        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
 
         def fix_output(o):
             """
@@ -310,7 +308,7 @@ class SonOfGridEngine(EngineBase):
     def queue_state(self):
         """Return s list with all currently running tasks in this queue"""
 
-        if time.time() - self._task_info_cache_last_update < gs.WAIT_PERIOD_BETWEEN_CHECKS:
+        if time.time() - self._task_info_cache_last_update < 30:
             # use cached value
             return self._task_info_cache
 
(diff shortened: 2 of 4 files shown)


albertz:
> If the SSH key isn't properly set, sisyphus will schedule stacked queue scan commands which will ask for a password

So you add `"-o", "ConnectTimeout"` to the SSH command? That seems like a bad workaround to me for this problem.

What about `-o BatchMode=yes`?

Or `-o PasswordAuthentication=no`?

albertz:
> ... sisyphus will schedule stacked queue scan commands which will ask for a password. This will lead to subprocess buildup, which eventually crashes the manager with a `Too many open files` error.

Also, I wonder about this. How can this happen? Are there uncleaned zombie procs? Did you check this? I think this is another separate thing you can and should fix.

Icemole:
> Btw, if we would just use `subprocess.run`, it would properly cleanup (kill) the subproc in case of a timeout. So instead of `p = subprocess.Popen(...)` and then `p.communicate`, just directly use `subprocess.run`.

I think that's a much more intuitive solution than what I proposed.

albertz:
> > Btw, if we would just use `subprocess.run`, it would properly cleanup (kill) the subproc in case of a timeout. So instead of `p = subprocess.Popen(...)` and then `p.communicate`, just directly use `subprocess.run`.
> 
> I think that's a much more intuitive solution than what I proposed.

Additionally, I still would add the `-o BatchMode=yes` or `-o PasswordAuthentication=no` though. There is no need to wait for the timeout.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,18 +86,23 @@ def system_call(self, command, send_to_stdin=None):
...
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
+        p = subprocess.run(
+            system_command,
+            stdin=send_to_stdin,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
+        )
```
albertz, line 105:
```suggestion
        p = subprocess.run(
            system_command,
            input=send_to_stdin,
            capture_output=True,
            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
        )
```

Icemole: Why were stdout/err a pipe to begin with? I thought that was needed by sisyphus :thinking:

albertz:
Btw, I wonder about similar wrong usages of `subprocess.Popen` in Sisyphus. There are probably a few. I just checked, it seems like in most of our backend engines, we have it wrong in the same way. Maybe you should apply the same fix for all of them?

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,16 +86,21 @@ def system_call(self, command, send_to_stdin=None):
...
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        p = subprocess.run(
+            system_command,
+            stdin=send_to_stdin,
+            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
+        )
```
albertz:
```suggestion
        p = subprocess.run(
            system_command,
            input=send_to_stdin,
            capture_output=True,
            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
        )
```

sisyphus/son_of_grid_engine.py
```diff
@@ -87,16 +87,21 @@ def system_call(self, command, send_to_stdin=None):
...
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        p = subprocess.run(
+            system_command,
+            stdin=send_to_stdin,
+            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
+        )
```
albertz:
```suggestion
        p = subprocess.run(
            system_command,
            input=send_to_stdin,
            capture_output=True,
            timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS,
        )
```

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,16 +86,17 @@ def system_call(self, command, send_to_stdin=None):
...
         else:
             # no gateway given, skip ssh local
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
```
albertz, line 99:
```suggestion
        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
```

sisyphus/son_of_grid_engine.py
```diff
@@ -87,16 +87,17 @@ def system_call(self, command, send_to_stdin=None):
...
         else:
             # no gateway given, skip ssh local
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        p = subprocess.run(system_command, stdin=send_to_stdin, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
```
albertz, line 100:
```suggestion
        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=gs.WAIT_PERIOD_BETWEEN_CHECKS)
```

Icemole:
> Btw, I wonder about similar wrong usages of subprocess.Popen in Sisyphus. There are probably a few. I just checked, it seems like in most of our backend engines, we have it wrong in the same way. Maybe you should apply the same fix for all of them?

I can do that as well. I actually envisioned joining that code into another common file because it's the exact same code, and we would be removing a bit of code duplication.

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: Because you need to check stdout/err afterwards? This must be there. But better use `capture_output=True` (shorter/simpler). But also, using `stdin` here is wrong. You need to use `input` instead.

albertz:
> > Btw, I wonder about similar wrong usages of subprocess.Popen in Sisyphus. There are probably a few. I just checked, it seems like in most of our backend engines, we have it wrong in the same way. Maybe you should apply the same fix for all of them?
> 
> I can do that as well.

I see the same wrong usage in these files:
* aws_batch_engine.py
* load_sharing_facility_engine.py
* simple_linux_utility_for_resource_management_engine.py
* son_of_grid_engine.py

> I actually envisioned joining that code into another common file because some of it is the exact same code, and we would be removing a bit of code duplication.

I would not do that for now. It's basically one line of code (`subprocess.run`). There is not really much code duplication. Moving this one line of code elsewhere would just make it more complicated. In any case, don't do this here in this PR.

critias:
> I would not do that for now. It's basically one line of code (`subprocess.run`). There is not really much code duplication. Moving this one line of code elsewhere would just make it more complicated. Or do you also mean the gateway logic? But in any case, I would not do this here in this PR.

I agree with Albert here.
And thanks for figuring this out, I didn't realized that the `subprocess.Popen` process would stay open.

Icemole:
Roger that :) will implement right away.

sisyphus/simple_linux_utility_for_resource_management_engine.py
Icemole: Okay, that makes sense. I think I've addressed all changes, resolving thread.

albertz:
Looks all fine now. Except maybe this aspect:

> I've also changed the hardcoded 30 second interval for an interval depending on `gs.WAIT_PERIOD_BETWEEN_CHECKS`, which I find makes more sense.

I wonder if that make sense to reuse this setting for the timeout. I personally would have kept those decoupled. I.e. for now, leave the 30 seconds, and if you want to be able to configure that, make a separate setting for it.

Icemole:
> Looks all fine now. Except maybe this aspect:
>
>>    I've also changed the hardcoded 30 second interval for an interval depending on gs.WAIT_PERIOD_BETWEEN_CHECKS, which I find makes more sense.
> 
> I wonder if that make sense to reuse this setting for the timeout. I personally would have kept those decoupled. I.e. for now, leave the 30 seconds, and if you want to be able to configure that, make a separate setting for it.

Okay, I'll change it back. I thought it made sense given the name.

For now I'll change it back to 30 seconds, but upon reviewing the `global_settings.py` file, I found the following potential candidates ("separate settings") already in there:

- ‎`‎WAIT_PERIOD_SSH_TIMEOUT`.
- `WAIT_PERIOD_QSTAT_PARSING`.

Icemole:
I've been testing this, and while my manager didn't crash anymore because of too many open files, there was an abnormally high memory usage that extended throughout time whenever the subprocesses didn't successfully finish:

```
[2024-06-13 07:24:26,842] INFO: Submit to queue: work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn run [1]
[2024-06-13 07:24:26,953] ERROR: Error to submit job, return value: 255
[2024-06-13 07:24:26,953] ERROR: SBATCH command: sbatch -J i6_core.returnn.extract_prior.ReturnnComputePriorJobV2.yYMFtGVTZgnn.run -o work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn/engine/%x.%A.%a [...] --wrap=/usr/bin/python3 /home/nbeneitez/work/sisyphus/too-many-open-files-fix/recipe/sisyphus/sis worker --engine long work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn run
[2024-06-13 07:24:26,953] ERROR: Error: [...] Permission denied (publickey).
```

When messages like this one were spammed, I noticed a very high memory usage from the manager's side. I'll try to review the code a bit more.

albertz (approved):

albertz:
The code looks fine to me though (sorry, forgot to review again, did that now).

You should debug this better. I.e. while you see the memory increase, is this memory increase really in the main proc of the manager? Or maybe just in some sub procs? Are the timed-out sub procs properly cleaned up (i.e. no zombie procs hanging around)?

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,16 +86,17 @@ def system_call(self, command, send_to_stdin=None):
...
         else:
             # no gateway given, skip ssh local
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
```
albertz, line 99: I wonder, don't you need to catch `TimeoutExpired`?

albertz:
I'm testing with a simple script whether there are any problems w.r.t. memory consumption or so. Script [here](https://github.com/albertz/playground/blob/master/py-subproc-timeout-cleanup.py), or just directly here:
```python
import subprocess

while True:
    try:
        subprocess.run(["cat", "/dev/zero"], timeout=0.01)
    except subprocess.TimeoutExpired:
        print("TimeoutExpired")
```
Just run this, and watch the memory consumption meanwhile. I don't see that there is any increase in memory.

So, I don't think that this is the problem.

albertz:
> When messages like this one were spammed

What do you mean by spammed? How often do you get them? It should wait each time for the timeout, or not?

sisyphus/simple_linux_utility_for_resource_management_engine.py
Icemole: Yes, you're right, I'll ignore any `TimeoutExpired`.

Icemole:
> You should debug this better. I.e. while you see the memory increase, is this memory increase really in the main proc of the manager? Or maybe just in some sub procs? Are the timed-out sub procs properly cleaned up (i.e. no zombie procs hanging around)?

Agreed, I panicked because I was making our head node go super slow and canceled the program :sweat_smile: I'll make a mental note to scan `/proc` thoroughly next time this happens.

> What do you mean by spammed? How often do you get them? It should wait each time for the timeout, or not?

It doesn't wait for the full timeout (30 seconds), but it's not spammy (as in < 1 second per print) either:
```
[2024-06-13 06:30:15,908] INFO: Submit to queue: work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn run [1]                                          
[2024-06-13 06:30:16,021] ERROR: Error to submit job, return value: 255     
...                                                                                               
[2024-06-13 06:30:28,284] INFO: Submit to queue: work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn run [1]                                          
```
The job immediately fails because of `Permission denied (publickey).`, and it seems to wait 10-20 seconds more before running another job.

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: So what does it mean? What happens now? The whole manager crashes?

Icemole: I don't think I had any `TimeoutExpired` exception, my error was within the timeout because of a public key mismatch, and it was actually outputted within the first second after the command was run:
```
[2024-06-13 07:24:26,842] INFO: Submit to queue: work/i6_core/returnn/extract_prior/ReturnnComputePriorJobV2.yYMFtGVTZgnn run [1]
[2024-06-13 07:24:26,953] ERROR: Error to submit job, return value: 255
```

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,16 +86,20 @@ def system_call(self, command, send_to_stdin=None):
...
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        try:
+            p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        except subprocess.TimeoutExpired:
+            logging.warning("Timeout expired for command: %s" % " ".join(system_command))
```
albertz, line 102: But this is the wrong way to handle this. Now `p` is undefined so you will get an exception later when `p` is accessed.

albertz:
```suggestion
            logging.warning("Timeout expired for command: %s" % " ".join(system_command))
            return [], ["TimeoutExpired"], -1
```
Or so?

Icemole:
I got into the issue again. So I scanned the process, and it seems to open many `submit_log.run` file descriptors at some point and immediately close them, and do nothing else for 15-30 seconds. The `ls` commands below were run with a timeout of 0.5 seconds more or less:

```
$ ls -lha /proc/3442960/fd
total 0
dr-x------ 2 nbeneitez domain_users  0 Jun 13 04:12 .
dr-xr-xr-x 9 nbeneitez domain_users  0 Jun 13 04:12 ..
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 0 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 1 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 10 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 11 -> 'pipe:[474983584]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 12 -> 'pipe:[474983584]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 13 -> 'pipe:[474967409]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 14 -> 'pipe:[474967409]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 15 -> 'pipe:[474988135]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:13 16 -> 'pipe:[474988135]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 2 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 3 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 4 -> 'socket:[474942638]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 5 -> 'socket:[474942639]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 6 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 7 -> 'socket:[474942644]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 8 -> 'socket:[474942645]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 9 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'

$ ls -lha /proc/3442960/fd
total 0
dr-x------ 2 nbeneitez domain_users  0 Jun 13 04:12 .
dr-xr-xr-x 9 nbeneitez domain_users  0 Jun 13 04:12 ..
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 0 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 1 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 10 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 11 -> 'pipe:[474983584]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 12 -> 'pipe:[474983584]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 13 -> 'pipe:[474967409]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 14 -> 'pipe:[474967409]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 15 -> 'pipe:[474988135]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:13 16 -> 'pipe:[474988135]'
lr-x------ 1 nbeneitez domain_users 64 Jun 14 06:06 17 -> 'pipe:[518288582]'
lr-x------ 1 nbeneitez domain_users 64 Jun 14 05:58 18 -> .../submit_log.run
lr-x------ 1 nbeneitez domain_users 64 Jun 14 06:06 19 -> .../submit_log.run
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 2 -> /dev/pts/8
lr-x------ 1 nbeneitez domain_users 64 Jun 14 05:43 20 -> .../submit_log.run
l-wx------ 1 nbeneitez domain_users 64 Jun 14 05:15 21 -> .../submit_log.run
lr-x------ 1 nbeneitez domain_users 64 Jun 14 05:21 22 -> .../submit_log.run
lr-x------ 1 nbeneitez domain_users 64 Jun 14 05:59 23 -> .../submit_log.run
lr-x------ 1 nbeneitez domain_users 64 Jun 14 05:21 24 -> .../submit_log.run
lr-x------ 1 nbeneitez domain_users 64 Jun 14 04:54 25 -> .../submit_log.run
l-wx------ 1 nbeneitez domain_users 64 Jun 14 05:52 26 -> 'pipe:[518288582]'
lr-x------ 1 nbeneitez domain_users 64 Jun 14 04:53 27 -> 'pipe:[518288583]'
l-wx------ 1 nbeneitez domain_users 64 Jun 14 06:16 28 -> 'pipe:[518288583]'
lr-x------ 1 nbeneitez domain_users 64 Jun 14 06:02 29 -> .../submit_log.run
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 3 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 4 -> 'socket:[474942638]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 5 -> 'socket:[474942639]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 6 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 7 -> 'socket:[474942644]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 8 -> 'socket:[474942645]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 9 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'

$ ls -lha /proc/3442960/fd
total 0
dr-x------ 2 nbeneitez domain_users  0 Jun 13 04:12 .
dr-xr-xr-x 9 nbeneitez domain_users  0 Jun 13 04:12 ..
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 0 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 1 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 10 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 11 -> 'pipe:[474983584]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 12 -> 'pipe:[474983584]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 13 -> 'pipe:[474967409]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:12 14 -> 'pipe:[474967409]'
lr-x------ 1 nbeneitez domain_users 64 Jun 13 04:12 15 -> 'pipe:[474988135]'
l-wx------ 1 nbeneitez domain_users 64 Jun 13 04:13 16 -> 'pipe:[474988135]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 2 -> /dev/pts/8
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 3 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 4 -> 'socket:[474942638]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 5 -> 'socket:[474942639]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 6 -> 'anon_inode:[eventpoll]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 7 -> 'socket:[474942644]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 8 -> 'socket:[474942645]'
lrwx------ 1 nbeneitez domain_users 64 Jun 13 04:12 9 -> '/dev/shm/pym-3442960-9n996ekj (deleted)'
```

All `submit_log.run` were pointing to different jobs. I don't see anything suspicious about this, it's just writing the queuing log to these files and then closing them. In general there seem to be 16 opened files, many of which are sockets and pipes.

The memory usage is right now at 6.4% of 250 GB and it keeps growing slowly but surely (update: 6.6% at the time of finishing the comment, so 10 minutes after more or less). `ps aux | grep <manager-pid>` only shows a process:

```
USER         PID %CPU %MEM      VSZ   RSS     TTY      STAT  START    TIME   COMMAND
nbeneit+ 3442960  164  6.5 22736408 17258504 pts/8 Sl+ Jun13 2585:55 python3 sis m config/training.py
```

but `htop` does show many processes with the same name but different PID, some of which have been running for some hours already. See the two images below.
imagen
imagen

In `proc/3442960/task` I see 106 subdirectories. So the program seems to spawn many threads, some of which aren't being killed, since according to `man 5 proc`:

> Underneath each of the /proc/pid directories, a task subdirectory contains subdirectories of the form task/tid, which contain corresponding information about each of the threads in the process, where tid is the kernel thread ID of the thread.

`subprocess.run` waits for the program to complete and then returns. Is it possible that some of the processes aren't being finished? But we should have the timeout for this :thinking: why are some subprocesses not being cleaned?

albertz:
Why do you actually think the memory leak is related to the PR here? It only increases when you get those "Error to submit job" and otherwise stays constant?

So are there alive subprocs or not? Are there any zombie procs? (What does `ps ax` say?)

What are the threads? (E.g. what does `pyspy` or `pystack` say?)

Where is the memory allocated? (What does a mem profiler say?)

Icemole:
Yes, it could very well be the case that there's some other error going on that's not related to the PR, but I was able to observe it after doing this change, so I've reported it here. If you want we can move this discussion to another PR.

There are no zombie processes. `ps ax | grep Z` reports nothing.

Sadly I don't have sudo powers and ptrace is not enabled, so I can't dump stats for `pystack`. However, there are exactly 106 warnings with the message `WARNING(process_remote): Failed to attach to thread <pid>: Operation not permitted`, with the PIDs being exactly the same as the ones in `/proc/3442960/task`. Besides, I've just noticed that all processes on `/proc/3442960/task` are a sequence order, which might point to the fact that they've been generated at the same time.

albertz:
> sudo powers and ptrace is not enabled

You don't need sudo for that. Just some admin has to enable ptrace. Then you can use it without ptrace. Please do that. I doesn't make sense that we waste time here by just guessing around without properly just debugging the issue.

Icemole:
Thanks for the suggestion, I've already asked to enable ptrace for everyone. In the meantime, one of our admins got me the trace we were looking for: [pystack_threads_tracker.log](https://github.com/user-attachments/files/15864055/pystack_threads_tracker.log).

From the log it would seem as if half of the threads (~50) were just starting up (see `task = get()`), and the other half were waiting for some job to finish. Moreover, the latter seem to be training-related since they go through `i6_core/returnn/training.py`. As you said, I indeed don't see any abnormal behavior here...

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: Actually, I realize now, in the code before your change, we also did not catch `TimeoutExpired` (in the `p.wait(timeout=30)`). I wonder why we did not catch it, and what happened when this exception was thrown there. Do you know? (As I understood you, you have run into this exact problem.)

Maybe we should not change this and also not catch `TimeoutExpired`? @critias?

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -86,16 +86,21 @@ def system_call(self, command, send_to_stdin=None):
...
             system_command = command
 
         logging.debug("shell_cmd: %s" % " ".join(system_command))
-        p = subprocess.Popen(system_command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        out, err = p.communicate(input=send_to_stdin, timeout=30)
+        try:
+            p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
+        except subprocess.TimeoutExpired:
```
michelwi, line 101: If we catch the `subprocess.TimeoutExpired` exception here and return, then the whole logic about retrying in `submit_helper` https://github.com/rwth-i6/sisyphus/blob/51db853857ab53b43ebb2f91b6e1347b81193bd7/sisyphus/simple_linux_utility_for_resource_management_engine.py#L239-L246 is obsolet. 
If this was intentional, then maybe also remove this logic? and `gs.WAIT_PERIOD_SSH_TIMEOUT` would be ignored then

albertz: Ah so it is actually handled. That was my earlier question about this. So then I would not catch it. Unless there is good reason to change this old behavior about `TimeoutExpired`.

Icemole: Yes... I hadn't noticed that it would be catched later in the call stack. See https://github.com/rwth-i6/sisyphus/pull/196.

PR #194, opened 2024-07-03
typing updates for Task

albertz:
Also removed Python 2 style code to derive from object.

I did not cover `update_rqmt` yet because this would require `Protocol` (not just `Callable`), and I'm not sure if we want that (requires Python 3.8).

### DIFF
--- a/sisyphus/task.py
+++ b/sisyphus/task.py
@@ -1,4 +1,3 @@
-from __future__ import annotations
 import math
 import os
 import logging
--- a/sisyphus/task.py
+++ b/sisyphus/task.py
@@ -1,9 +1,10 @@
+from __future__ import annotations
 import math
 import os
 import logging
 import sys
 import time
-import typing
+from typing import Optional, Union, Any, Sequence, Dict, List
 import subprocess as sp
 from ast import literal_eval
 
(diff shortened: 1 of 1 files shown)


sisyphus/task.py
```diff
@@ -3,35 +3,35 @@
...
         self,
-        start,
-        resume=None,
-        rqmt=None,
-        args=None,
-        mini_task=False,
+        start: str,
+        resume: Optional[str] = None,
+        rqmt: Optional[Dict[str, Any]] = None,
+        args: Optional[Sequence[Union[List[Any], Any]]] = None,
```
michelwi, line 24:
```suggestion
        args: Optional[Sequence[Any]] = None,
```
can already be `Any` and `List[Any]` is contained in `Any`.

michelwi:
>  did not cover update_rqmt yet because this would require Protocol (not just Callable), and I'm not sure if we want that

Why would `Callable[[Dict[str, Any], Dict[str, Any]], Dict[str, Any]]` not work?

sisyphus/task.py
albertz: Yea I know, but sometimes I prefer this to give some more specific type hints to what types would be specifically expected here, or specially handled, even though other types (any other types) would also be ok.

Also, I basically just translated the typing from the earlier docstring. I also wondered a bit about why this is here at all, and I thought there is indeed maybe a special meaning when there is a list?

albertz:
> > did not cover update_rqmt yet because this would require Protocol (not just Callable), and I'm not sure if we want that
> 
> Why would `Callable[[Dict[str, Any], Dict[str, Any]], Dict[str, Any]]` not work?

We call `self._update_rqmt(last_rqmt=last_rqmt, last_usage=last_usage)`, i.e. we use named args, which is not supported by `Callable`.

michelwi (approved):

NeoLegends (approved):
👍

PR #196, opened 2024-07-03
Catch `subprocess.TimeoutError` once

Icemole:
This was the root cause of several jobs with the same features being submitted, even though they had already been submitted:

1. We use the `gateway` option and there's a `TimeoutError` [here](https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/simple_linux_utility_for_resource_management_engine.py#L101).
2. From the point of view of sisyphus, the `ssh gw-02 squeue` finishes with error -1. Sisyphus then reaches [here](https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/simple_linux_utility_for_resource_management_engine.py#L241) safely. It hasn't crashed with a `TimeoutError` because we haven't propagated it.
3. No `TimeoutError` is raised after calling `self.system_call()` because it's already been addressed inside `self.system_call()` and it hasn't been propagated. Therefore, the process ignores the `except` block [here](https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/simple_linux_utility_for_resource_management_engine.py#L242). If `retval == -1`, sisyphus doesn't care either!

In my view, the problem comes from not addressing errors that could have been obtained from the return values. Therefore, this PR correctly addresses return codes different from zero, and leaves the work of addressing the exceptions of the subprocesses to `self.system_call()`.

### DIFF
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -90,11 +90,8 @@ class AWSBatchEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        try:
-            p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
-        except subprocess.TimeoutExpired:
-            logging.warning("Timeout expired for command: %s" % " ".join(system_command))
-            return [], ["TimeoutExpired".encode()], -1
+
+        p = subprocess.run(system_command, input=send_to_stdin, capture_output=True, timeout=30)
 
         def fix_output(o):
             """
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -58,11 +58,8 @@ class LoadSharingFacilityEngine(EngineBase):
         logging.debug("shell_cmd: %s" % " ".join(system_command))
         if send_to_stdin:
             send_to_stdin = send_to_stdin.encode()
-        try:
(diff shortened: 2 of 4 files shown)


albertz:
For reference, related is the recent change in PR #191 on how to handle `TimeoutExpired `.

albertz:
> We use the `gateway` option and there's a `TimeoutError` [here](https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/simple_linux_utility_for_resource_management_engine.py#L101).

I think you mean `TimeoutExpired`?

albertz:
I already asked in #191 whether catching `TimeoutExpired` is really correct, or whether we just should pass on this exception, exactly as you do now in this PR. In #191, there was no answer to that. See https://github.com/rwth-i6/sisyphus/pull/191#discussion_r1642474520 and https://github.com/rwth-i6/sisyphus/pull/191#discussion_r1644426773.

albertz:
As I wrote before, I wonder, why do we actually do this change? Why not keep the original behavior, i.e. just not handle `TimeoutExpired` in `system_call`?

albertz:
> I will therefore leave the code as it was, catching the `TimeoutExpired` outside `system_call`.

I.e. you update this PR here, to remove the `try:`/`except TimeoutExpired:` in `system_call`?

Icemole:
> > I will therefore leave the code as it was, catching the TimeoutExpired outside system_call.
> 
> I.e. you update this PR here, to remove the try:/except TimeoutExpired: in system_call?

Yes! I did that just now :)

albertz (approved):

PR #198, opened 2024-07-17
Queue scan: handle return value different from zero

### LINKED ISSUE #175: More then one matching SLURM task (again)
Previous (but all probably independent issues): #163, #156

```
[2024-01-24 20:36:01,045] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.mxzabPJ0zKAb.run', 1) > matches: [('4297533', 'PENDING'), ('4296236', 'RUNNING')]                   
[2024-01-24 20:36:01,046] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.y94krGqVhAjX.run', 1) > matches: [('4296231', 'RUNNING'), ('4297531', 'RUNNING')]  
[2024-01-24 20:36:01,053] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.g41HLN8rZS2S.run', 1) > matches: [('4297534', 'PENDING'), ('4296233', 'RUNNING')]    
...
[2024-01-24 20:47:09,468] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.mxzabPJ0zKAb.run', 1) > matches: [('4297533', 'PENDING'), ('4296236', 'RUNNING')]
...
[2024-01-24 20:48:29,415] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.mxzabPJ0zKAb.run', 1) > matches: [('4297533', 'PENDING'), ('4296236', 'RUNNING')]  
[2024-01-24 20:48:29,420] WARNING: More then one matching SLURM task, use first match < ('i6_core.returnn.forward.ReturnnForwardJobV2.g41HLN8rZS2S.run', 1) > matches: [('4297534', 'PENDING'), ('4296233', 'RUNNING')]                   
...
```
Unfortunately I don't see the previous log of Sisyphus anymore, so no idea how this could have happened.

Icemole:
Fix #175.

Whenever the return value of the queue scan command is different from zero, the job states are eventually set as `UNKNOWN`. Whenever some job is set as `UNKNOWN` and there's no log file attached, the job will be automatically considered as `RUNNABLE` and queued.

This is a very dangerous process because there's usually a job already queued "naturally", that is, queued because the queue scan job finished gracefully, found nothing, and set the job state as `UNKNOWN`, thus queuing the job normally.

This wrong process can happen many times, until the first job scheduled eventually enters `RUNNING` state, and thus a log file is generated, preventing any further same job schedulings.

This PR fixes such wrong behavior by making the queue scan wait some seconds before rescheduling it again after a failure in the scan, and not directly setting the job state as `UNKNOWN`.

### DIFF
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -235,7 +235,7 @@ class AWSBatchEngine(EngineBase):
         if qs == "FAILED":
             return STATE_QUEUE_ERROR
         logging.warning("Unknown AWS engine state %s" % qs)
-        return STATE_UNKNOWN
+        return STATE_QUEUE_ERROR
 
     def start_engine(self):
         """No starting action required with the current implementation"""
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -26,7 +26,7 @@ from collections import namedtuple
 
 import sisyphus.global_settings as gs
 from sisyphus.engine import EngineBase
-from sisyphus.global_settings import STATE_RUNNABLE, STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
+from sisyphus.global_settings import STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
 
 ENGINE_NAME = "aws"
 TaskInfo = namedtuple("TaskInfo", ["job_id", "task_id", "state"])
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -26,7 +26,7 @@ from collections import namedtuple
 
 import sisyphus.global_settings as gs
 from sisyphus.engine import EngineBase
-from sisyphus.global_settings import STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
+from sisyphus.global_settings import STATE_RUNNABLE, STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
 
 ENGINE_NAME = "aws"
 TaskInfo = namedtuple("TaskInfo", ["job_id", "task_id", "state"])
@@ -225,7 +225,7 @@ class AWSBatchEngine(EngineBase):
         try:
             queue_state = self.queue_state()
         except subprocess.CalledProcessError:
-            return STATE_QUEUE_ERROR
+            return STATE_RUNNABLE
         qs = queue_state.get(task_name)
 
         # task name should be uniq
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -222,7 +222,10 @@ class AWSBatchEngine(EngineBase):
         """Return task state:"""
         name = task.task_name()
         task_name = escape_name(name, task_id)
-        queue_state = self.queue_state()
+        try:
+            queue_state = self.queue_state()
+        except subprocess.CalledProcessError:
+            return STATE_QUEUE_ERROR
         qs = queue_state.get(task_name)
 
         # task name should be uniq
--- a/sisyphus/aws_batch_engine.py
+++ b/sisyphus/aws_batch_engine.py
@@ -222,10 +222,7 @@ class AWSBatchEngine(EngineBase):
         """Return task state:"""
         name = task.task_name()
         task_name = escape_name(name, task_id)
-        try:
-            queue_state = self.queue_state()
-        except subprocess.CalledProcessError:
-            return STATE_RUNNABLE
+        queue_state = self.queue_state()
         qs = queue_state.get(task_name)
 
         # task name should be uniq
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -11,7 +11,7 @@ from collections import defaultdict, namedtuple
 
 import sisyphus.global_settings as gs
 from sisyphus.engine import EngineBase
-from sisyphus.global_settings import STATE_RUNNABLE, STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
+from sisyphus.global_settings import STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
 
 ENGINE_NAME = "lsf"
 TaskInfo = namedtuple("TaskInfo", ["job_id", "task_id", "state"])
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -195,7 +195,7 @@ class LoadSharingFacilityEngine(EngineBase):
                 out, err, retval = self.system_call(bsub_call, command)
                 if retval != 0:
                     logging.warning(self._system_call_error_warn_msg(bsub_call))
-                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
                     continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(bsub_call))
@@ -260,7 +260,7 @@ class LoadSharingFacilityEngine(EngineBase):
                 out, err, retval = self.system_call(system_command)
                 if retval != 0:
                     logging.warning(self._system_call_error_warn_msg(system_command))
-                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
                     continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -193,10 +193,6 @@ class LoadSharingFacilityEngine(EngineBase):
             logging.info("command: %s" % command)
             try:
                 out, err, retval = self.system_call(bsub_call, command)
-                if retval != 0:
-                    logging.warning(self._system_call_error_warn_msg(bsub_call))
-                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
-                    continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(bsub_call))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -255,9 +255,7 @@ class LoadSharingFacilityEngine(EngineBase):
             try:
                 out, err, retval = self.system_call(system_command)
                 if retval != 0:
-                    logging.warning(self._system_call_error_warn_msg(system_command))
-                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
-                    continue
+                    raise subprocess.CalledProcessError(self._system_call_error_warn_msg(system_command))
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
@@ -292,7 +290,10 @@ class LoadSharingFacilityEngine(EngineBase):
         name = task.task_name()
         name = escape_name(name).encode()
         task_name = (name, task_id)
-        queue_state = self.queue_state()
+        try:
+            queue_state = self.queue_state()
+        except subprocess.CalledProcessError:
+            return STATE_QUEUE_ERROR
         qs = queue_state[task_name]
 
         # task name should be uniq
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -255,7 +255,7 @@ class LoadSharingFacilityEngine(EngineBase):
             try:
                 out, err, retval = self.system_call(system_command)
                 if retval != 0:
-                    raise subprocess.CalledProcessError(self._system_call_error_warn_msg(system_command))
+                    raise subprocess.CalledProcessError(retval, self._system_call_error_warn_msg(system_command))
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -48,6 +48,11 @@ class LoadSharingFacilityEngine(EngineBase):
             return f"SSH command timeout: {command!s}"
         return f"Command timeout: {command!s}"
 
+    def _system_call_error_warn_msg(self, command: Any) -> str:
+        if self.gateway:
+            return f"SSH command error: {command!s}"
+        return f"Command error: {command!s}"
+
     def system_call(self, command, send_to_stdin=None):
         if self.gateway:
             system_command = ["ssh", "-x", self.gateway] + [" ".join(["cd", os.getcwd(), "&&"] + command)]
@@ -188,6 +193,10 @@ class LoadSharingFacilityEngine(EngineBase):
             logging.info("command: %s" % command)
             try:
                 out, err, retval = self.system_call(bsub_call, command)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(bsub_call))
+                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(bsub_call))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
@@ -249,6 +258,10 @@ class LoadSharingFacilityEngine(EngineBase):
         while True:
             try:
                 out, err, retval = self.system_call(system_command)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(system_command))
+                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -255,9 +255,9 @@ class LoadSharingFacilityEngine(EngineBase):
             try:
                 out, err, retval = self.system_call(system_command)
                 if retval != 0:
-                    raise subprocess.CalledProcessError(
-                        retval, system_command, self._system_call_error_warn_msg(system_command)
-                    )
+                    logging.warning(self._system_call_error_warn_msg(system_command))
+                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
+                    continue
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
@@ -292,10 +292,7 @@ class LoadSharingFacilityEngine(EngineBase):
         name = task.task_name()
         name = escape_name(name).encode()
         task_name = (name, task_id)
-        try:
-            queue_state = self.queue_state()
-        except subprocess.CalledProcessError:
-            return STATE_RUNNABLE
+        queue_state = self.queue_state()
         qs = queue_state[task_name]
 
         # task name should be uniq
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -255,7 +255,9 @@ class LoadSharingFacilityEngine(EngineBase):
             try:
                 out, err, retval = self.system_call(system_command)
                 if retval != 0:
-                    raise subprocess.CalledProcessError(retval, self._system_call_error_warn_msg(system_command))
+                    raise subprocess.CalledProcessError(
+                        retval, system_command, self._system_call_error_warn_msg(system_command)
+                    )
             except subprocess.TimeoutExpired:
                 logging.warning(self._system_call_timeout_warn_msg(system_command))
                 time.sleep(gs.WAIT_PERIOD_SSH_TIMEOUT)
--- a/sisyphus/load_sharing_facility_engine.py
+++ b/sisyphus/load_sharing_facility_engine.py
@@ -11,7 +11,7 @@ from collections import defaultdict, namedtuple
 
 import sisyphus.global_settings as gs
 from sisyphus.engine import EngineBase
-from sisyphus.global_settings import STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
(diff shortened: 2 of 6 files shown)


sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -235,6 +240,10 @@ def submit_helper(self, call, logpath, rqmt, name, task_name, start_id, end_id,
         while True:
             try:
                 out, err, retval = self.system_call(sbatch_call)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(sbatch_call))
+                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    continue
```
michelwi, line 246: I don't think we should do this. There is already code below that handles submission errors

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -307,6 +316,10 @@ def queue_state(self):
         while True:
             try:
                 out, err, retval = self.system_call(system_command)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(system_command))
+                    time.sleep(gs.WAIT_PERIOD_BETWEEN_CHECKS)
+                    continue
```
michelwi, line 322: Not sure if this solves the problem.. If there is some fundamental problem with the queue, that is not recoverable, then this sends the function into an endless loop.

maybe instead just return the old cache?
```suggestion
                if retval != 0:
                    return self._task_info_cache
```
In the worst case we will never update the cache and a running/finished Job will still be marked as pending.

michelwi (changes requested):

albertz:
Is this related to #175? Or is there any issue which describes the problem?

sisyphus/load_sharing_facility_engine.py
```diff
@@ -48,6 +48,11 @@ def _system_call_timeout_warn_msg(self, command: Any) -> str:
             return f"SSH command timeout: {command!s}"
         return f"Command timeout: {command!s}"
 
+    def _system_call_error_warn_msg(self, command: Any) -> str:
```
christophmluscher, line 51: Is this maybe a method which could go into `EngineBase`?

sisyphus/simple_linux_utility_for_resource_management_engine.py
christophmluscher: IMHO I would log and wait, so that the normal cache update process can run

michelwi:
> Is this related to #175? 

yes, possibly

sisyphus/simple_linux_utility_for_resource_management_engine.py
christophmluscher: define a new variable?

sisyphus/simple_linux_utility_for_resource_management_engine.py
Icemole: Technically, the snippet right below handles timeouts, not errors. Both the `logging.warning` and the `time.sleep` are called with different arguments.

sisyphus/load_sharing_facility_engine.py
Icemole: Maybe, but applying the same train of thought we'd also have to migrate `_system_call_timeout_warn_msg`. Your call.

sisyphus/son_of_grid_engine.py
```diff
@@ -74,11 +73,6 @@ def __init__(self, default_rqmt, gateway=None, auto_clean_eqw=True, ignore_jobs=
         self.ignore_jobs = ignore_jobs
         self.pe_name = pe_name
 
-    def _system_call_timeout_warn_msg(self, command: Any) -> str:
-        if self.gateway:
-            return f"SSH command timeout: {command!s}"
-        return f"Command timeout: {command!s}"
-
```
michelwi, line 81: Is this code not used below?

sisyphus/simple_linux_utility_for_resource_management_engine.py
michelwi: I was referring to https://github.com/rwth-i6/sisyphus/blob/ce5f7a2cf70ceac7cf330fdb6e7ab3578892ef8c/sisyphus/simple_linux_utility_for_resource_management_engine.py#L248-L285

albertz:
> Whenever the return value of the queue scan command is different from zero, the job states are eventually set as `UNKNOWN`. Whenever some job is set as `UNKNOWN` and there's no log file attached, the job will be automatically considered as `RUNNABLE` and queued.

What do you mean by "queue scan command"? You mean what happens in the `queue_state` function? Or in the `submit_call` function?

What do you mean by "log file attached"?

> This is a very dangerous process because there's usually a job already queued "naturally", that is, queued because the queue scan job finished gracefully, found nothing, and set the job state as `UNKNOWN`, thus queuing the job normally.

What do you mean by "finished gracefully"? Above you said the return value is different from zero? So it's not finished gracefully then? Or is it?

What do you mean by "there's usually a job already queued "naturally", that is, queued because the queue scan job finished gracefully"? Why is the job already queued? If the command exists with non-zero code, usually the job is not queued, or not?

What do you mean by "queue scan job"? Before you said "queue scan command". Do you mean the same? Do you actually mean the `queue_state` function?

> This wrong process can happen many times, until the first job scheduled eventually enters `RUNNING` state, and thus a log file is generated, preventing any further same job schedulings.
> 
> This PR fixes such wrong behavior by making the queue scan wait some seconds before rescheduling it again after a failure in the scan, and not directly setting the job state as `UNKNOWN`.

What do you mean by "scan"?

Sorry, I don't fully understand the actual problem. Aren't you saying there is a bug in `queue_state`? But then, why not fix this bug? It seems like instead of fixing the bug, you add a workaround at another place? But maybe I just didn't fully understand it.

Icemole:
@albertz I'll answer you in line.

>> Whenever the return value of the queue scan command is different from zero, the job states are eventually set as UNKNOWN. Whenever some job is set as UNKNOWN and there's no log file attached, the job will be automatically considered as RUNNABLE and queued.
> 
> What do you mean by "queue scan command"? You mean what happens in the queue_state function? Or in the submit_call function?

Yes, I mean what happens in the `queue_state` function. More specifically, I mean running the actual `squeue`, `qstat`, etc commands. If you have standard nomenclature for this, please let me know to avoid any further confusion.

> What do you mean by "log file attached"?

I mean the log file attached to the job/task pair. For instance, if we're running some job whose task is `run`, by "log file attached" I mean the `log.run.1` file created on disk. You can also see this as the normal log file generated in the filesystem by starting the job/task pair.

>What do you mean by "queue scan job"? Before you said "queue scan command". Do you mean the same? Do you actually mean the queue_state function?

Yes, I mean the same, apologies for the confusing nomenclature. I mean running `squeue`, `qstat`, etc.

>>This wrong process can happen many times, until the first job scheduled eventually enters RUNNING state, and thus a log file is generated, preventing any further same job schedulings.
>>This PR fixes such wrong behavior by making the queue scan wait some seconds before rescheduling it again after a failure in the scan, and not directly setting the job state as UNKNOWN.
>
> What do you mean by "scan"?

Running `sbatch`, `qstat`, etc. I mean the "queue scan [command/job]".

>>This is a very dangerous process because there's usually a job already queued "naturally", that is, queued because the queue scan job finished gracefully, found nothing, and set the job state as UNKNOWN, thus queuing the job normally.
>
>What do you mean by "finished gracefully"? Above you said the return value is different from zero? So it's not finished gracefully then? Or is it?
>
>What do you mean by "there's usually a job already queued "naturally", that is, queued because the queue scan job finished gracefully"? Why is the job already queued? If the command exists with non-zero code, usually the job is not queued, or not?
>
> Sorry, I don't fully understand the actual problem. Aren't you saying there is a bug in queue_state? But then, why not fix this bug? It seems like instead of fixing the bug, you add a workaround at another place? But maybe I just didn't fully understand it.

In my original comment I was referring to two situations that interact wrongly among each other, let me detail them further (assume SLURM is the manager):

1. The job/task pair predecessor finishes.
2. The queue is scanned for the job/task pair. The `squeue` command is executed, and it returns 0 (successful, "the queue scan job finished gracefully") and there's no job/task pair waiting in the queue. Moreover, there's no `log.<task>.X` file ("log file attached") in the filesystem. As a consequence, the job state becomes `RUNNABLE`, and is thus sent into the queue shortly via `sbatch`. Note that, if there had been any `log.<task>.X` file, the job state would have become `UNKNOWN` instead, because of how `Job._is_runnable()` works.
3. In a minute or so, another `squeue` command is executed, while the previous job is still on queue. This last part is key for the issue we're seeing, because the `log.<task>.X` won't be generated. Take this into account when reading below.
4. This time the `squeue` command fails with whatever error code, let's assume -1, not relevant. With the code as it previously was, there was no distinction for any return value different from zero when submitting a `squeue` command, so the manager assumes there's some stdout (empty) and some stderr (probably with some text), but the manager is only interested in the stdout. After parsing it (empty), it realizes that there's no job/task pair waiting in the queue (which is wrong). Moreover, there's no `log.<task>.X` file ("log file attached") in the filesystem, as per (3). As a consequence, the job/task pair is detected as runnable, and is thus sent into the queue shortly via `sbatch`.

Note that the wrong behavior from the unintended behavior of "same job queued twice" comes from the `sbatch` command failing to be rerun whenever its return code is different from zero, as stated above. Therefore, I believe my proposed code isn't a workaround at all, but an actual solution to our issue.

I hope I have clarified your doubts. Please let me know if there are still any questions or comments.

sisyphus/simple_linux_utility_for_resource_management_engine.py
Icemole: Got it, thanks for the comment. I've now removed all references of return value checks in all actual queue submit commands.

albertz:
> This time the `squeue` command fails with whatever error code, let's assume -1, not relevant. With the code as it previously was, there was no distinction for any return value different from zero when submitting a `squeue` command

What do you mean "the code as it previously was"? The error code?

What do you mean "no distinction for any return value different from zero"? I fail to parse the grammar of this sentence. The `squeue` command has a return value (error code) different from 0, as you said, or not?

I'm looking at `queue_state` now. One problem seems to be that we parse the stdout no matter of the return value (`retval`).

But ok. It returns an empty `defaultdict(list)` then. The only place `queue_state` is used is in the `task_state` function. It should run into this code:
```python
queue_state = self.queue_state()
qs = queue_state[task_name]
if qs == []:  # <--- this should be True
    return STATE_UNKNOWN
```
So it will return `STATE_UNKNOWN`. But that is different from `STATE_RUNNABLE`?

albertz:
Ah, in `Task._get_state_helper`:
```python
if engine_state == gs.STATE_UNKNOWN:
    if self.started(task_id):
        ...
    else:
        return gs.STATE_RUNNABLE
```
`started` will return False when the log file does not exist.
And that is what you get in `Task.state()` then. So that is why it thinks this task is runnable.

But isn't that the main problem here? This logic I just showed from `Task._get_state_helper`, that `STATE_UNKNOWN` and log file does not exist -> `STATE_RUNNABLE`? Shouldn't we change/fix that? In what cases does this logic make sense?

michelwi:
> But isn't that the main problem here? This logic I just showed from `Task._get_state_helper`, that `STATE_UNKNOWN` and log file does not exist -> `STATE_RUNNABLE`? Shouldn't we change/fix that? In what cases does this logic make sense?

This logic is used in the case when a job is (actually) runnable: `STATE_UNKNOWN` in the engine means "This job is not known to the engine". There could be three reasons for it:
1. it has not been submitted yet. This is what sisyphus assumes is the case here and (re-)submits the job
2. the squeue call failed. This is not handled in the master branch and likely causes the problem
3. the job is finished. In this case the worker has set the finished marker and `_get_state_helper` will return STATE_FINISHED [here](https://github.com/rwth-i6/sisyphus/blob/ce5f7a2cf70ceac7cf330fdb6e7ab3578892ef8c/sisyphus/task.py#L393)

To fix the problem we should fix 2. I tend to agree with Albert that 
> One problem seems to be that we parse the stdout no matter of the return value (retval).

In SGE the `_task_info_cache` is not updated when there is a [parsing error](https://github.com/rwth-i6/sisyphus/blob/ce5f7a2cf70ceac7cf330fdb6e7ab3578892ef8c/sisyphus/son_of_grid_engine.py#L332) maybe we should do something similar here as well?

As an aside: I noticed that, when a job is submitted, it is [added to the](https://github.com/rwth-i6/sisyphus/blob/ce5f7a2cf70ceac7cf330fdb6e7ab3578892ef8c/sisyphus/simple_linux_utility_for_resource_management_engine.py#L265) `_task_info_cache` as PENDING and even when a [reset of the cache](https://github.com/rwth-i6/sisyphus/blob/ce5f7a2cf70ceac7cf330fdb6e7ab3578892ef8c/sisyphus/simple_linux_utility_for_resource_management_engine.py#L287) is requested, it is not actually dropped, only the next call to `queue_state` will trigger a call to squeu.

So if we avoid setting the `_task_info_cache` to an empty dict, we should be good.

albertz:
> This logic is used in the case when a job is (actually) runnable: `STATE_UNKNOWN` in the engine means "This job is not known to the engine".

If that is really what `STATE_UNKNOWN` means, then it means, `task_state()` is wrong, and we should fix that.

michelwi:
> If that is really what `STATE_UNKNOWN` means, 

That would be my interpretation based on the code.

> then it means, `task_state()` is wrong, and we should fix that.

how would you propose to fix that?

albertz:
> > then it means, `task_state()` is wrong, and we should fix that.
> 
> how would you propose to fix that?

Maybe introduce a new state `STATE_QUEUE_ERROR`  or so, which tells the engine that the job is in an unknown state.

Oh, I just saw, there is already such a state. I think `queue_state` should throw an exception if `retval != 0`, and we catch that exception in `task_state`, and return `STATE_QUEUE_ERROR` if we get this.

Icemole:
Thanks for the comments. I've added `STATE_QUEUE_ERROR` just like in the rest of the engines. Let me know if this is what you intended.

albertz:
> I've added `STATE_QUEUE_ERROR` just like in the rest of the engines.

But you still have this wrong code there?
```python
if qs == []:
    return STATE_UNKNOWN
```

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -358,7 +367,7 @@ def task_state(self, task, task_id):
         elif state in ["PENDING", "CONFIGURING"]:
             return STATE_QUEUE
         else:
-            return STATE_UNKNOWN
+            return STATE_QUEUE_ERROR
```
michelwi, line 370: No, that's wrong. This else is the case where a task is legitimately not found in the queue (i.e. finished or not yet submitted). The queue error would be in the ` qs == []` case. 

Although I don't think we can arrive there any more with the addition of `if retval != 0: continue`

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -307,6 +312,10 @@ def queue_state(self):
         while True:
             try:
                 out, err, retval = self.system_call(system_command)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(system_command))
+                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
+                    continue
```
albertz, line 318: But this here is not what was discussed? What we discussed should be sth like this here:
```suggestion
                    raise subprocess.CalledProcessError(self._system_call_error_warn_msg(system_command))
```
Together with the other changes in `task_state`, as it was discussed.

sisyphus/load_sharing_facility_engine.py
```diff
@@ -249,6 +254,10 @@ def queue_state(self):
         while True:
             try:
                 out, err, retval = self.system_call(system_command)
+                if retval != 0:
+                    logging.warning(self._system_call_error_warn_msg(system_command))
+                    time.sleep(gs.WAIT_PERIOD_QSTAT_PARSING)
+                    continue
```
albertz, line 260: Same here.

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: > The queue error would be in the ` qs == []` case.

See my other comment. We should not check for `qs == []`. We should use a `try: ... except CalledProcessError: return STATE_QUEUE_ERROR`.

I think the `return STATE_QUEUE_ERROR` here is fine. We only get here if we have an unexpected state. That doesn't mean that the engine does not know about the job. It knows about the job but fails to parse the state.

michelwi: > I think the `return STATE_QUEUE_ERROR` here is fine. We only get here if we have an unexpected state. That doesn't mean that the engine does not know about the job. It knows about the job but fails to parse the state.

I beg to differ. I think we arrive here when the job has not yet been submitted and with this change we cannot submit any job.

albertz: But you have this code above:
```python
queue_state = self.queue_state()
qs = queue_state[task_name]
```
When `qs` is empty, that means that there are no corresponding tasks reported by `squeue` (and thus it should return `STATE_UNKNOWN`, which means exactly that: as you said, "this job is not known to the engine").
When `qs` is non-empty, it means there *are* corresponding tasks reported by `squeue`. So that by your definition implies that it cannot return `STATE_UNKNOWN`, because it is known to the engine. If the state is not one of the expected states, it means the task is known but something unexpected happened, i.e. that's exactly what `STATE_QUEUE_ERROR` means.

michelwi: Yes, sorry, I had assumed the empty list would result from a faulty `squeue` call. But in fact it is just what we get from the defaultdict if a key is missing. 

I take back the original comment

curufinwe:
Imo we should not throw an exception if the parsing of the queue output fails, but treat it in the same way as a timeout error and assume that the error is transient and will go away (either by itself or through admin intervention). Thus we should just keep on trying until we get a correct output from squeue.

sisyphus/load_sharing_facility_engine.py
Icemole: I will resolve this. Please reopen it if you think it's necessary. As stated, I would just leave it for another PR.

Icemole:
> we should not throw an exception if the parsing of the queue output fails, but treat it in the same way as a timeout error and assume that the error is transient and will go away (either by itself or through admin intervention).

That's more or less what we were doing before, and what we're doing now with the new functionality as described by @albertz.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -307,6 +312,8 @@ def queue_state(self):
         while True:
             try:
                 out, err, retval = self.system_call(system_command)
+                if retval != 0:
+                    raise subprocess.CalledProcessError(retval, self._system_call_error_warn_msg(system_command))
```
michelwi, line 316: `subprocess.CalledProcessError` needs the returncode and the command
```suggestion
                    raise subprocess.CalledProcessError(retval, system_command, self._system_call_error_warn_msg(system_command))
```
or define your own Exception.

michelwi (changes requested):
> > we should not throw an exception if the parsing of the queue output fails, but treat it in the same way as a timeout error and assume that the error is transient and will go away (either by itself or through admin intervention).
> 
> That's more or less what we were doing before, and what we're doing now with the new functionality as described by @albertz.

Yes, but it is not retried very often. I tested it now and found 
```
[2024-08-19 10:09:39,353] INFO: Finished updating job states
[2024-08-19 10:09:39,367] INFO: Experiment directory: /path/to/my/setup_folder     Call: /path/to/sisyphus/sis m config/my_config.py
[2024-08-19 10:09:39,387] INFO: queue_error(11) waiting(72)
Print verbose overview (v), update aliases and outputs (u), start manager (y), or exit (n)? y
[2024-08-19 10:10:38,269] INFO: There is nothing I can do, good bye!
```
after retrying once (after 1 min) the manager terminates

Icemole:
> after retrying once (after 1 min) the manager terminates

I would definitely always retry. This seems to be a rather small issue. Should the `STATE_QUEUE_ERROR` be redefined for this purpose?

michelwi:
> > after retrying once (after 1 min) the manager terminates
> 
> I would definitely always retry. This seems to be a rather small issue. Should the `STATE_QUEUE_ERROR` be redefined for this purpose?

I think this would need be added in the manager [work_left](https://github.com/rwth-i6/sisyphus/blob/4ca86d2a53761041b5aed62055d1c722a5eda440/sisyphus/manager.py#L326) function. But not sure about other side effects.

Icemole:
We're forgetting that `QUEUE_ERROR` represents a job in `Eqw` (SGE equivalent) state. Maybe we should do as Eugen suggested and make it a `RUNNABLE` task instead of having it fail with `QUEUE_ERROR`.

Icemole:
So we can't use `STATE_RUNNABLE` because there's an assertion when getting the state of a job:

```
File ~/work/sisyphus/too-many-open-files-fix/recipe/sisyphus/sisyphus/task.py:371, in Task._get_state_helper(self=<Task 'run' job=Job<work/i6_core/recognition/conversion/LatticeToCtmJob.Cy7LtqTAYoNa>>, engine=<sisyphus.engine.EngineSelector object>, task_id=1)
    369 else:
    370     engine_state = engine.task_state(self, task_id)
--> 371     assert engine_state in (gs.STATE_QUEUE, gs.STATE_QUEUE_ERROR, gs.STATE_RUNNING, gs.STATE_UNKNOWN)
```

---

I think implementing Eugen's plan would be best: treat the wrong queue command right after it fails, just like a timeout error. There are some alternatives:

1. Returning to a previous version from this same PR in which that was done: see [here](https://github.com/rwth-i6/sisyphus/pull/198/commits/f4e2f1a75d0931db4cce7d3eecdd88c333d132bb).
2. In a solution that might make everyone more comfortable, the code could be something like:

```
def task_state(self, task, task_id):
    # ...
    while True:
        try:
            queue_state = self.queue_state()
            break
        except subprocess.CalledProcessError:
            pass
```

Note: we can add some functionality to make the code do that, say, 10 times, and actually enter `STATE_QUEUE_ERROR` if it can't be scheduled in that amount of tries.

As a reminder, the current code is:

```
def task_state(self, task, task_id):
    # ...
    try:
        queue_state = self.queue_state()
    except subprocess.CalledProcessError:
        return STATE_QUEUE_ERROR  # or STATE_RUNNABLE
```

which is wrong. `STATE_QUEUE_ERROR` is meant to be used for a job in `Eqw` state. And I don't feel very comfortable modifying code inside the `Engine` class so that the queue state accepts `STATE_RUNNABLE` as well just for this.

michelwi:
> STATE_QUEUE_ERROR is meant to be used for a job in Eqw state.

ok.

> and make it a RUNNABLE task instead

that is wrong as well. A runnable task will be (re-)submitted by the manager. This is exactly the behavior before the PR that we wanted to fix.

>Thus we should just keep on trying until we get a correct output from squeue.

I agree here. My proposal how to do this would be similar to how sge does it
https://github.com/rwth-i6/sisyphus/blob/4ca86d2a53761041b5aed62055d1c722a5eda440/sisyphus/son_of_grid_engine.py#L329-L342
If there is an error, just retry. This might (probably) also block the manager from submitting while the problem persists.

Icemole:
I think this is done. Please check if you like this approach.

sisyphus/aws_batch_engine.py
```diff
@@ -26,7 +26,7 @@
 
 import sisyphus.global_settings as gs
 from sisyphus.engine import EngineBase
-from sisyphus.global_settings import STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
+from sisyphus.global_settings import STATE_RUNNABLE, STATE_RUNNING, STATE_UNKNOWN, STATE_QUEUE, STATE_QUEUE_ERROR
```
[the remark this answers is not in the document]
curufinwe: Ruff agrees and I fixed it here and everywhere else.

sisyphus/simple_linux_utility_for_resource_management_engine.py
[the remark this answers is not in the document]
curufinwe: done

curufinwe (approved):

michelwi (approved):
Tested this version on slurm and it works as expected.

PR #199, opened 2024-07-17
allow keeping input folder during job cleanup

vieting:
When sisyphus cleans up a job folder, the input folder is deleted. This can be impractical if you want to find some job's dependencies because instead of just going through the input folders, you need to open each `finished.tar.gz` file and check for input jobs in the `info` file. I therefore propose to add the config option `JOB_CLEANUP_KEEP_INPUT`. It set to True, the input folder will remain even after cleaning.

Edit: As suggested in the PR, we set `JOB_CLEANUP_KEEP_INPUT = True` as the default setting. This changes the default behavior, but is not expected to hurt anyone.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -133,7 +133,7 @@ JOB_CLEANER_WORKER = 5
 #: If the job internal work directory should be kept or deleted during clean up
 JOB_CLEANUP_KEEP_WORK = False
 #: If the job internal input directory with symlinks to input jobs should be kept or deleted during clean up
-JOB_CLEANUP_KEEP_INPUT = False
+JOB_CLEANUP_KEEP_INPUT = True
 #: Default value for job used by tk.cleaner to determine if a job should be removed or not
 JOB_DEFAULT_KEEP_VALUE = 50
 #: How many threads should update the graph in parallel, useful if the filesystem has a high latency
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -130,8 +130,10 @@ JOB_AUTO_CLEANUP = True
 JOB_CLEANER_INTERVAL = 60
 #: How many threads should be cleaning in parallel
 JOB_CLEANER_WORKER = 5
-#: If the job internal work directory should be keeped re deleted during clean up
+#: If the job internal work directory should be kept or deleted during clean up
 JOB_CLEANUP_KEEP_WORK = False
+#: If the job internal input directory with symlinks to input jobs should be kept or deleted during clean up
+JOB_CLEANUP_KEEP_INPUT = False
 #: Default value for job used by tk.cleaner to determine if a job should be removed or not
 JOB_DEFAULT_KEEP_VALUE = 50
 #: How many threads should update the graph in parallel, useful if the filesystem has a high latency
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -516,13 +516,10 @@ class Job(metaclass=JobSingleton):
                 try:
                     if not gs.JOB_CLEANUP_KEEP_WORK:
                         shutil.rmtree(os.path.abspath(self._sis_path(gs.JOB_WORK_DIR)))
-                    if not gs.JOB_CLEANUP_KEEP_INPUT:
-                        shutil.rmtree(os.path.abspath(self._sis_path(gs.JOB_INPUT)))
-                    files = [
-                        i
-                        for i in os.listdir(self._sis_path())
-                        if i not in (gs.JOB_OUTPUT, gs.JOB_INFO, gs.JOB_WORK_DIR, gs.JOB_INPUT)
-                    ]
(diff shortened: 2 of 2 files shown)


michelwi (approved):
This changes the default behavior from "add input folder to tar archive" to "delete input folder". But I even slightly prefer the new behavior.

vieting:
> This changes the default behavior from "add input folder to tar archive" to "delete input folder". But I even slightly prefer the new behavior.

Right, should have mentioned this in the description. I'd expect that an input folder in the tar archive is hardly ever used. But we can also keep it in there if others prefer that.

albertz:
> you need to open each `finished.tar.gz` file and check for input jobs in the `info` file

I'm not sure I fully understand. You don't need to open the `finished.tar.gz`. You can just open the `info` file directly, which is not removed?

sisyphus/job.py
```diff
@@ -516,10 +516,12 @@ def _sis_cleanup(self):
                 try:
                     if not gs.JOB_CLEANUP_KEEP_WORK:
                         shutil.rmtree(os.path.abspath(self._sis_path(gs.JOB_WORK_DIR)))
+                    if not gs.JOB_CLEANUP_KEEP_INPUT:
```
Atticus1806, line 519: I don't like this to be honest. I have had cases in the past where I would unzip the cleaned file to just have the input folder. While I will most likely set the flag, I feel like this can potentially cause spots where information is a lot harder to get (if the info file is weird/bloated).

sisyphus/global_settings.py
```diff
@@ -130,8 +130,10 @@ def file_caching(path):
 JOB_CLEANER_INTERVAL = 60
 #: How many threads should be cleaning in parallel
 JOB_CLEANER_WORKER = 5
-#: If the job internal work directory should be keeped re deleted during clean up
+#: If the job internal work directory should be kept or deleted during clean up
 JOB_CLEANUP_KEEP_WORK = False
+#: If the job internal input directory with symlinks to input jobs should be kept or deleted during clean up
+JOB_CLEANUP_KEEP_INPUT = False
```
Atticus1806, line 136: I would even vote for having this be enabled by default (if this breaks nothing)

Atticus1806 (changes requested):

sisyphus/global_settings.py
albertz: What happens if `input` is kept and is also in the `finished.tar.gz` and then you unzip the `finished.tar.gz`? Will that lead to an error because the dir `input` and its file already exists?

sisyphus/job.py
albertz: I don't exactly understand the case that the `info` file is weird/bloated. Can you give some example? In my scripts, I just always extract this information automatically from the `info` file, and never had any problems.

Atticus1806: In a RASR recognition usually the whole RASR config is in the info file. If you have split your data into mulitple hdfs then each of them is also in there, even though they might point to the same job. 
I just personally prefer to be able to cd into the dir (and then easily cd back with ..).

sisyphus/global_settings.py
Atticus1806: But it should not be in the zipped file if its kept right?

albertz: Currently it is never in the zipped file, even if it is not kept. I thought this was supposed to be changed. I'm not sure how it is then. So you say it should not be in the zipped file if it is kept?

Atticus1806: Yes in this PR you are right. But I would change this behavior to only be included in the zip if the flag is not set. I guess my other comment kind of included that to me, but I did not make this clear.

sisyphus/job.py
```diff
@@ -516,10 +516,12 @@ def _sis_cleanup(self):
...
+                        shutil.rmtree(os.path.abspath(self._sis_path(gs.JOB_INPUT)))
                     files = [
                         i
                         for i in os.listdir(self._sis_path())
-                        if i not in (gs.JOB_OUTPUT, gs.JOB_INFO, gs.JOB_WORK_DIR)
+                        if i not in (gs.JOB_OUTPUT, gs.JOB_INFO, gs.JOB_WORK_DIR, gs.JOB_INPUT)
```
Atticus1806, line 524: If we don't delete the input of course zip it here if its not going to be kept outside.

vieting:
> > you need to open each `finished.tar.gz` file and check for input jobs in the `info` file
> 
> I'm not sure I fully understand. You don't need to open the `finished.tar.gz`. You can just open the `info` file directly, which is not removed?

Right, the `info` file is not inside the tar archive. Still, I prefer to easily step through the input dependencies by going through the input folders. This is way easier than checking the input in the `info` file, then going to that job's `info` file and so forth.

Atticus1806 (approved):

michelwi (approved):

PR #203, opened 2024-08-01
add "report" section to webserver

michelwi:
This adds a new tab to the sisyphus webserver to display the `OutputReports`.


NeoLegends (commented):
As discussed offline, we are thinking about adding a separate type of report that is known to produce HTML to solve the problem of whether to HTML-escape the report output, or not, without relying on heuristics.

PR #205, opened 2024-08-19
Allow hooking job failure for generic error handling

### LINKED ISSUE #179: Auto-restart jobs on user-specified error conditions
There are certain errors which we frequently see in our cluster, where I want that my job in error state is automatically restarted.

Examples:

* `RuntimeError: cuFFT error: CUFFT_INTERNAL_ERROR`
* (other similar CUDA errors)
* CUDA OOM or CPU OOM, *and* it trained at least one epoch since last run

In that case, currently I just manually delete the `error` and `submit_log`.

I want some way in the Sisyphus that the manager can automatically do that for me.

How?

I'm not sure. Maybe some custom user callback once a job gets into error state? One suggestion was made in #204:
```python
def task_exit(
    *,
    job: tk.Job,
    task_name: str,
    success: bool,
    engine_info: Any,  # last engine_info from submit_log
    usage: Any,  # usage from usage log
    call: List[str],  # actual call after worker_wrapper
    # log: str,  # could also feed in the log output here? depending on the job very memory intensive...
) -> Optional[bool]: ...
```

Additionally, sometimes I want to temporarily add some cluster nodes to the exclude list, maybe for 1 day or so, when they fail again and again with the same CUDA error. This should also be possible somehow.

### LINKED ISSUE #204: Hooking job exit/failure
Sometimes, when e.g. a job fails due to some local disk running full on one cluster node, I wish I was able to detect this condition in the manager and reschedule the job on a different node w/o manual intervention.

It would be nice if sisyphus allowed hooking job failure/success in some way externally through `gs`, in a similar fashion to how `worker_wrapper` works. I could imagine something like:

```python
import os
from typing import Any, List, Optional
from sisyphus import tk


def task_exit(
    *,
    job: tk.Job,
    task_name: str,
    success: bool,
    engine_info: Any,  # last engine_info from submit_log
    usage: Any,  # usage from usage log
    call: List[str],  # actual call after worker_wrapper
    # log: str,  # could also feed in the log output here? depending on the job very memory intensive...
) -> Optional[bool]:
    if success:
        return

    with open(os.path.join(job.work_path(), f"log.{task_name}.1")) as log_file:
        for line in log_file:
            if "We cannot free enough space on /ssd" in line:
                job.update_rqmt(
                    {"sbatch_args": [*job.rqmt.sbatch_args, "-x", usage["host"]]}
                )
                return True  # reschedule
```

NeoLegends:
Closes #179
Closes #204

now testing this

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -67,10 +67,15 @@ def on_job_failure(job):
     Can be used for generic job-independent error monitoring, handling or retry
     logic.
 
-    Sispyhus will call this function w/ the job instance if the job enters the
-    failure state. The callback itself is then responsible for any retry logic,
-    realized by e.g. analyzing the job log file and removing error files in the job
-    directory as needed.
+    Sispyhus will call this function w/ the job instance for any failed job.
+
+    The callback itself is then responsible for any retry logic, realized by e.g.
+    analyzing the job log file and removing error files in the job directory as
+    needed.
+
+    The callback needs to be stateless and indempotent, as it can be called multiple
+    times on the same job, especially if the job remains in the error state after the
+    callback has finished.
 
     Do:
         - use with caution
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -60,6 +60,26 @@ def worker_wrapper(job, task_name, call):
     return call
 
 
+def on_job_failure(job):
+    """
+    Job failure hook.
+
+    Can be used for generic job-independent error monitoring, handling or retry
+    logic.
+
+    Sispyhus will call this function w/ the job instance if the job enters the
+    failure state. The callback itself is then responsible for any retry logic,
+    realized by e.g. analyzing the job log file and removing error files in the job
+    directory as needed.
+
+    Do:
+        - use with caution
+        - ensure you don't build infinite retry loops
+        - limit to specific use cases (e.g. local disk full, GPU broken, etc.)
+    """
+    pass
+
+
 def update_engine_rqmt(last_rqmt: Dict, last_usage: Dict):
     """Update requirements after a job got interrupted, double limits if needed
 
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -576,12 +576,6 @@ class Manager(threading.Thread):
             self.job_cleaner.start()
         return True
 
-    def handle_job_failure(self, prev_jobs: Dict[str, List["Job"]], cur_jobs: Dict[str, List["Job"]]):
-        prev_errored_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
-        for job in cur_jobs.get(gs.STATE_ERROR, []):
-            if job not in prev_errored_jobs:
-                gs.on_job_failure(job)
-
     @tools.default_handle_exception_interrupt_main_thread
     def run(self):
         if not self.startup():
@@ -604,8 +598,7 @@ class Manager(threading.Thread):
 
             config_manager.continue_readers()
 
-            prev_jobs = self.jobs
-            cur_jobs = self.update_jobs()
+            self.update_jobs()
 
             if gs.CLEAR_ERROR or self.clear_errors_once:
                 self.clear_errors_once = False
@@ -631,7 +624,9 @@ class Manager(threading.Thread):
             self.setup_holded_jobs()
             self.resume_jobs()
             self.run_jobs()
-            self.handle_job_failure(prev_jobs, cur_jobs)
+
+            for job in self.jobs.get(gs.STATE_ERROR, []):
+                gs.on_job_failure(job)
 
         # Stop config reader
         config_manager.cancel_all_reader()
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -5,6 +5,7 @@ import os
 import sys
 import threading
 import time
+from typing import TYPE_CHECKING, Dict, List
 import warnings
 
 from multiprocessing.pool import ThreadPool
@@ -14,6 +15,9 @@ from sisyphus.loader import config_manager
 from sisyphus.block import Block
 from sisyphus.tools import finished_results_cache
 import sisyphus.global_settings as gs
+
+if TYPE_CHECKING:
+    from sisyphus.job import Job
 
 
 class JobCleaner(threading.Thread):
@@ -572,6 +576,12 @@ class Manager(threading.Thread):
             self.job_cleaner.start()
         return True
 
+    def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
+        prev_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
+        for job in cur_jobs.get(gs.STATE_ERROR, []):
+            if job not in prev_jobs:
+                gs.on_job_failure(job)
+
     @tools.default_handle_exception_interrupt_main_thread
     def run(self):
         if not self.startup():
@@ -593,7 +603,9 @@ class Manager(threading.Thread):
             self.check_output(write_output=self.link_outputs)
 
             config_manager.continue_readers()
-            self.update_jobs()
+
+            prev_jobs = self.jobs
+            cur_jobs = self.update_jobs()
 
             if gs.CLEAR_ERROR or self.clear_errors_once:
                 self.clear_errors_once = False
@@ -619,6 +631,7 @@ class Manager(threading.Thread):
             self.setup_holded_jobs()
             self.resume_jobs()
             self.run_jobs()
+            self.handle_job_failure(prev_jobs, cur_jobs)
 
         # Stop config reader
         config_manager.cancel_all_reader()
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -577,9 +577,9 @@ class Manager(threading.Thread):
         return True
 
     def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
-        prev_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
+        prev_errored_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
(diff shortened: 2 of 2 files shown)


sisyphus/manager.py
```diff
@@ -572,6 +576,12 @@ def maybe_clear_state(state, always_clear, action):
             self.job_cleaner.start()
         return True
 
+    def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
+        prev_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
+        for job in cur_jobs.get(gs.STATE_ERROR, []):
+            if job not in prev_jobs:
```
critias, line 582: Is it possible that this line should be `if job not in prev_jobs.get(gs.STATE_ERROR, []):`?

sisyphus/manager.py
```diff
@@ -572,6 +576,12 @@ def maybe_clear_state(state, always_clear, action):
             self.job_cleaner.start()
         return True
 
+    def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
```
critias, line 579: Please add a short doc string here as well.

critias (commented):
Add another doc string, but beside that it looks good to me.

About the proposed usage in the linked issues: It might be worth having a more general way detect broken nodes and exclude them for a certain time instead of adding this to each job individually 🤔

sisyphus/global_settings.py
```diff
@@ -60,6 +60,26 @@ def worker_wrapper(job, task_name, call):
...
 
+def on_job_failure(job):
+    """
+    Job failure hook.
+
+    Can be used for generic job-independent error monitoring, handling or retry
+    logic.
+
+    Sispyhus will call this function w/ the job instance if the job enters the
+    failure state. The callback itself is then responsible for any retry logic,
```
michelwi, line 71:
```suggestion
    error state. The callback itself is then responsible for any retry logic,
```
I think it is called "error" everywhere else. cf. also "interrupted_resumable" and "interrupted_non_resumable" which I would also name failures but they are not handled here.

Maybe the function name should be renamed as well.

sisyphus/manager.py
```diff
@@ -593,7 +603,9 @@ def run(self):
             self.check_output(write_output=self.link_outputs)
 
             config_manager.continue_readers()
-            self.update_jobs()
+
+            prev_jobs = self.jobs
+            cur_jobs = self.update_jobs()
```
michelwi, line 608: How will this interact with the block directly below. `self.clear_states` removes some errors, call `self.update_jobs()`  and then cycle the outer loop again. This would mess up the logic to detect new jobs in error state.

sisyphus/manager.py
```diff
@@ -572,6 +576,12 @@ def maybe_clear_state(state, always_clear, action):
             self.job_cleaner.start()
         return True
 
+    def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
+        prev_errored_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
+        for job in cur_jobs.get(gs.STATE_ERROR, []):
+            if job not in prev_errored_jobs:
+                gs.on_job_failure(job)
+
```
michelwi, line 584: would you call `self.update_jobs()` after finishing the loop to account for changes in the state of jobs?

sisyphus/manager.py
```diff
@@ -572,6 +576,12 @@ def maybe_clear_state(state, always_clear, action):
             self.job_cleaner.start()
         return True
 
+    def handle_job_failure(self, prev_jobs: Dict[str, List[Job]], cur_jobs: Dict[str, List[Job]]):
+        prev_errored_jobs = set(prev_jobs.get(gs.STATE_ERROR, []))
+        for job in cur_jobs.get(gs.STATE_ERROR, []):
+            if job not in prev_errored_jobs:
```
michelwi, line 582: do we need to keep track of which job is newly added to error state or could this be a stateless function and we always use it on all errored jobs?

sisyphus/manager.py
NeoLegends: Not needed, since this will loop around immediately when the processing is done.

sisyphus/manager.py
```diff
@@ -593,6 +597,7 @@ def run(self):
             self.check_output(write_output=self.link_outputs)
 
             config_manager.continue_readers()
+
```
michelwi, line 600:
```suggestion
```

michelwi (commented):
I guess some changes can now be removed again

michelwi (approved):

NeoLegends:
As a first approximation of a proper error handling implementation for CUDA errors I'm encountering at the momentadding this snippet to `settings.py` works reasonably well:

```python
ignore_error_cache = set()

def on_job_failure(job: "Job"):
    from i6_core.returnn import ReturnnTrainingJob

    if not isinstance(job, ReturnnTrainingJob):
        logging.debug(f"{job.job_id()}: error, but not a {ReturnnTrainingJob.__name__}, so not doing anything.")
        return
    elif job.job_id() in ignore_error_cache:
        return

    log_file_path = os.path.join(job.work_path(), "../log.run.1")
    with open(log_file_path, "rt") as log_file:
        is_cuda_err = any("cuda error" in line.lower() for line in log_file)

    if not is_cuda_err:
        logging.debug(f"{job.job_id()}: died but probably not due to a CUDA error, better go check by hand.")
        ignore_error_cache.add(job.job_id())
        return

    logging.info(f"{job.job_id()}: CUDA 💥, re-starting... 🔁")

    # archive log file
    i = 1
    cleared_log_path = None
    while cleared_log_path is None or os.path.exists(cleared_log_path):
        cleared_log_path = os.path.join(job.work_path(), f"../log.run.cuda-cleared.{i:04}.gz")
        i += 1
    with open(log_file_path, "rb") as log_in, gzip.open(cleared_log_path, "wb") as log_out:
        shutil.copyfileobj(log_in, log_out)
    os.remove(log_file_path)

    # re-schedule job
    for f in [
        os.path.join(job.work_path(), "../error.run.1"),
        os.path.join(job.work_path(), "../submit_log.run"),
    ]:
        try:
            os.remove(f)
        except FileNotFoundError:
            pass
```

albertz:
I wonder, this callback is called all the time, not once on job failure? This is a bit unexpected to me. And also makes the logic much more complicated on the user side. E.g. you need to add this `ignore_error_cache` logic here. Which is also wrong, because once the user clears the error for this job, and it continues to run, it might later run into a CUDA error, but then you would ignore it, because you never clear the `ignore_error_cache` here.

albertz:
Well, ok, checking the mtime of the error file probably should be better, if you want to keep the callback logic this way. Like:
```python

ignore_error_cache = {}  # job_id -> err_mtime


# https://github.com/rwth-i6/sisyphus/pull/205#issuecomment-2368527715
def on_job_failure(job: Job):
    import logging
    import gzip
    from i6_core.returnn import ReturnnTrainingJob

    if not isinstance(job, ReturnnTrainingJob):
        return

    err_mtime = os.path.getmtime(os.path.join(job.work_path(), "../error.run.1"))
    if ignore_error_cache.get(job.job_id()) == err_mtime:
        return

    log_file_path = os.path.join(job.work_path(), "../log.run.1")
    with open(log_file_path, "rt") as log_file:
        is_cuda_err = any(("cuda error" in line.lower() or "cuFFT error" in line) for line in log_file)

    if not is_cuda_err:
        logging.debug(f"{job.job_id()}: died but probably not due to a CUDA error, better go check by hand.")
        ignore_error_cache[job.job_id()] = err_mtime
        return

    ...
```

NeoLegends:
Version that works with multi-node trainings:

```python
ignore_error_cache = set()


def on_job_failure(job: "Job"):
    import glob
    import gzip
    import logging

    from i6_core.returnn import ReturnnTrainingJob

    if not isinstance(job, ReturnnTrainingJob):
        logging.debug(f"{job.job_id()}: error, but not a {ReturnnTrainingJob.__name__}, so not doing anything.")
        return

    try:
        err_mtime = os.path.getmtime(os.path.join(job.work_path(), "../error.run.1"))
    except FileNotFoundError:
        return  # maybe was already cleared

    if ignore_error_cache.get(job.job_id(), None) == err_mtime:
        return

    log_files = [f for f in glob.iglob(os.path.join(job.work_path(), "../log.run.*")) if not f.endswith("gz")]

    is_cuda_err = False
    for log_file_path in log_files:
        try:
            with open(log_file_path, "rt") as log_file:
                is_cuda_err = any(
                    "cuda error" in line_l or "cufft error" in line_l
                    for line in log_file
                    for line_l in [line.casefold()]
                )
        except Exception as exc:
            logging.warning(
                f"{job.job_id()}: died, but could not open the log file {log_file_path} for some reason: {exc}"
            )
        if is_cuda_err:
            break

    if not is_cuda_err:
        logging.warning(f"{job.job_id()}: died but probably not due to a CUDA error, better go check by hand.")
        ignore_error_cache[job.job_id()] = err_mtime
        return

    logging.info(f"{job.job_id()}: CUDA 💥, re-starting... 🔁")

    # archive log file(s)
    i = 1
    cleared_log_path = None
    while cleared_log_path is None or os.path.exists(cleared_log_path):
        cleared_log_path = os.path.join(job.work_path(), f"../log.run.cuda-cleared.{i:04}.gz")
        i += 1
    with gzip.open(cleared_log_path, "wb") as log_out:
        for log_file_path in log_files:
            with open(log_file_path, "rb") as log_in:
                shutil.copyfileobj(log_in, log_out)
            os.remove(log_file_path)

    # re-schedule job
    for f in [
        *glob.iglob(os.path.join(job.work_path(), "../error.run.*")),
        os.path.join(job.work_path(), "../submit_log.run"),
    ]:
        try:
            os.remove(f)
        except FileNotFoundError:
            pass
```

albertz:
For reference, I use this now:
```python
def on_job_failure(job: Job):
    from i6_experiments.users.zeyer.sis_tools import job_failure_handler

    return job_failure_handler.get_cached_job_failure_handler()(job)
```

Issue #207, opened 2024-10-08
Hash (`get_object_state`) of `functools.partial` is wrong


albertz:
Current `get_object_state` logic:

```python
    if hasattr(obj, "__getnewargs_ex__"):
        args = obj.__getnewargs_ex__()
    elif hasattr(obj, "__getnewargs__"):
        args = obj.__getnewargs__()
    else:
        args = None

    if hasattr(obj, "__sis_state__"):
        state = obj.__sis_state__()
    elif hasattr(obj, "__getstate__"):
        state = obj.__getstate__()
    elif hasattr(obj, "__dict__"):
        state = obj.__dict__
    elif hasattr(obj, "__slots__"):
        state = {k: getattr(obj, k) for k in obj.__slots__ if hasattr(obj, k)}
    else:
        assert args is not None, "Failed to get object state of: %s" % repr(obj)
        state = None
```

I think this is already partially wrong when there is both `__dict__` and `__slots__` (which is valid).

Also note that the behavior changed since Python 3.11: Now there is always a default `__getstate__` function ([doc](https://docs.python.org/3/library/pickle.html#object.__getstate__)), which actually handles `__dict__`/`__slots__` correctly, but that means, the behavior changed for Sisyphus hashes from older Python versions to Python 3.11 and newer (but only for those rare cases where there was both `__dict__` and `__slots__`).

`functools.partial` is such a case:
```python
class partial:
    __slots__ = "func", "args", "keywords", "__dict__", "__weakref__"
    ...
```
Note that the `__dict__` usually stays empty. It's just there to allow to add other attribs to the object (which would not be possible if there is no `__dict__`).

So, for older Python versions, it's wrong.

For Python >=3.11, you might think, it's correct.

But it's also wrong, now in a different way! After the pure Python implementation of `partial`, it tries to import a native implementation to replace it, as an optimization:
```python
try:
    from _functools import partial
except ImportError:
    pass
```
This might succeed or not, depending on Python compiler flags (it could be optional to compile this native `_functools` module).

Now, this native `partial` is more tricky. It again defines `__dict__`. But it does not have slots. Instead, the relevant args are stored internally in the native `partial` type. So, just accessing `__dict__` here is wrong, but also using `__getstate__` is wrong, as this will just return `__dict__` (it's the default implementation of `__getstate__`).

(I'm not really sure whether that's maybe even a bug in CPython. It was unexpected to me. I reported it here: https://github.com/python/cpython/issues/125094)

[...]

PR #208, opened 2024-10-08
hash get_object_state, use dir for native types, fix __slots__ 

albertz:
Fix https://github.com/rwth-i6/sisyphus/issues/207

See #207 for some discussion.

Note that this slightly changes the behavior, but only for cases which were basically broken before.

E.g., for `functools.partial`, before this change, none of the relevant args (`func`, `args`, `keywords`) were ever part of the hash. So this change should not really break anything.

The same is also true for cases where there was both a `__dict__` and `__slots__`. With Python <=3.10, it would have used only the `__dict__`, and very likely, that was just empty, and does not contain any of the relevant things (which are very likely all in the `__slots__`), so it was totally broken and basically unusable. With Python >=3.11, there is a default `__getstate__` now, which would already do the correct thing, at least for "normal" (non-native) types, i.e. merge `__dict__` and `__slots__`, i.e. our existing `get_object_state` was correct for those cases. With this change, Python <=3.10 is also correct now.

In any case, we check through `dir` now, and we check for member descriptors, which should also cover such native types (like `functools.partial`).

### DIFF
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -159,23 +159,20 @@ def _getmembers(obj):
         if key in res:
             continue
         # Get class attribute first, to maybe skip descriptors.
-        has_cls_attr = False
-        cls_value = None
         if key in cls_dict:
-            has_cls_attr = True
             cls_value = cls_dict[key]
             if hasattr(cls_value, "__get__"):  # descriptor
+                # descriptor are e.g. properties, bound methods, etc. We don't want to have those.
+                # But member descriptors are usually for slots (even for native types without __slots__),
+                # so that is why we keep them.
                 if not ismemberdescriptor(cls_value):
-                    # descriptor are e.g. properties, bound methods, etc. We don't want to have those.
-                    # but not member d
                     continue
-            print("  -", key, cls_value)
         try:
             value = getattr(obj, key)
         except AttributeError:
             # dir might not be reliable. just skip this
             continue
-        if has_cls_attr and cls_value is value:
+        if key in cls_dict and cls_dict[key] is value:
             continue  # this is a class attribute
         res[key] = value
     return res
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -1,6 +1,6 @@
 import enum
 import hashlib
-from inspect import isclass, isfunction
+from inspect import isclass, isfunction, ismemberdescriptor
 
 
 def md5(obj):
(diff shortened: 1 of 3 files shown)


albertz:
Btw, I just noticed, our `extract_paths` has the same problem. I actually wonder, why does it not use `get_object_state`, but instead its own logic? Should I also fix this in this PR here, or make a separate PR for this (I think I would just use `get_object_state` there).

NeoLegends:
So, it seems like w/ the new implementation `pathlib.Path` hashes differently. Old sis gave: 
```
>>> p = pathlib.Path("/home/mgunz/src/apptek_asr/[...]/test-hash.py")
>>> from sisyphus.hash import get_object_state
>>> get_object_state(p)
{}
```

New sis gives:
```
>>> p = pathlib.Path("/home/mgunz/src/apptek_asr/[...]/test-hash.py")
>>> from sisyphus.hash import get_object_state
>>> get_object_state(p)
{'_drv': '', '_parts': ['/', 'home', 'mgunz', 'src', 'apptek_asr', ..., 'test-hash.py'], '_root': '/'}
```

The reason for this is that if you instantiate a `pathlib.Path`, on unix systems you get
```
class PosixPath(Path, PurePosixPath):
    """Path subclass for non-Windows systems.

    On a POSIX system, instantiating a Path should return this object.
    """
    __slots__ = ()
```

Notice the empty slots. Before this PR, sisyphus would therefore not take into account _any_ properties for the hash. `get_object_state(pathlib.Path)` would give you `(PosixPath, (dict))`. Now, w/ the changes you instead get `PosixPath, (dict, (tuple, (str, '_cached_cparts'), (list, (str, '/'), (str, 'nas'), (str, 'data'), (str, 'speech'), ...)), (tuple, (str, '_drv'), (str, '')), (tuple, (str, '_parts'), (list, (str, '/'), (str, 'nas'), (str, 'data'), (str, 'speech'), ...)), (tuple, (str, '_root'), (str, '/'))))`

albertz:
Ah, `PosixPath` seems to behave somewhat bad in both cases though (although some state is obviously better than no state at all). The state you get there with the new code doesn't really seem to be so stable across Python versions.

But I also wonder, the `__slots__` is always empty? That means, all the state should be in `__dict__`? So then the old code should have worked already, or not?

Btw, also please say what Python version you used. For the new code, the Python version should not make a difference (well, despite if the object changed its internal representation - that might also be the case for `PosixPath`, I don't know). For the old code, it will also be different depending on Python version. With Python >=3.11, I think the state was not empty for `PosixPath`. Maybe check that as well.

In any case, maybe it makes sense to treat `PosixPath` as a special case anyway, and just store its path as state? (But that would be a separate PR then.)

Also, I wonder, is that correct that you have `PosixPath` as argument here? Maybe you actually want to have a Sisyphus `Path` instead?

curufinwe:
In general, yes. A sisyphus Path would be preferable, but we should make sure that sisyphus also works correctly with PosixPath. I would not mind handling it separately, but then we should do it in this PR to avoid having another commit than changes hashes.

albertz:
Ok, I added special case for `pathlib.PurePath`.

sisyphus/hash.py
```diff
@@ -43,6 +44,13 @@ def get_object_state(obj):
     Comment: Maybe obj.__reduce__() is a better idea? is it stable for hashing?
     """
 
+    if isinstance(obj, pathlib.PurePath):
+        # pathlib paths have a somewhat technical internal state
+        # ('_drv', '_root', '_parts', '_str', '_hash', '_pparts', '_cached_cparts'),
+        # so we don't want to rely on this, but instead just use the string representation as state.
+        # https://github.com/rwth-i6/sisyphus/pull/208#issuecomment-2405560718
+        return str(obj)
```
michelwi, line 52: Does now `pathlib.Path("foo")` get the same hash as `"foo"`? 
Would that be a problem?

albertz: No. See the added test case.

NeoLegends (approved):

curufinwe (approved):

PR #211, opened 2024-10-22
 Optimize job pickling to speed up Manager thread performance on large graphs

RaphaelLilt:
**Description**
We found that the manager threads were starting new jobs deep in a large graph slowly (approximately one job every 5 seconds on average). When profiling with `py-spy`, we found that the pickling in `sisyphus.job.Job._sis_setup_directory` was a performance bottleneck. Additionally, the pickled jobs files (`job.save`) were very large—reaching up to 104 MB for jobs that only had file paths as input—which is as we found out since they contain all their preprocessor jobs (e.g. `job.input_files[0].creator.input_files[0].creator...`).

**Proposed Changes**
- Modify `sisyphus.job_path.AbstractPath.__getstate__` used by Pickle to set the `creator` to `None` and adding a method `__sis_state__` used by `sisyphus.hash.get_object_state`. The worker running a job only needs to know the jobs input paths but not their creators.
- This ensures that the hashes don't change and the pickled jobs only contain the necessary state.
- In `global_constants.py` add `INCLUDE_CREATOR_STATE` to control the inclusion of `"creator"` fields in serialized `AbstractPath`.

**Results:**
- **Increased Speed:** After the change we were able to simultaneously run 10,000 jobs compared to approximately 5 Jobs before for large graphs.
- **Reduced Pickled File Size**: The pickled job files are much smaller about 100 KB compared to the previous > 104 MB.

This optimization leads to much faster job initializations for large graphs making the system to scale to larger graphs.

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -205,6 +205,9 @@ ALIAS_AND_OUTPUT_SUBDIR = ""
 #: Show job targets on status screen, can significantly slow down startup time if many outputs are used
 SHOW_JOB_TARGETS = True
 
+# Controls whether to include 'creator' in serialized state of an abstract path
+INCLUDE_CREATOR_STATE = True
+
 #: How many seconds should be waited before assuming a job is finished after the finished file is written
 #: to allow network file system to sync up
 WAIT_PERIOD_JOB_FS_SYNC = 30
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -206,7 +206,7 @@ ALIAS_AND_OUTPUT_SUBDIR = ""
 SHOW_JOB_TARGETS = True
 
 # Controls whether to include 'creator' in serialized state of an abstract path
-INCLUDE_CREATOR_STATE = True
+INCLUDE_CREATOR_STATE = False
 
 #: How many seconds should be waited before assuming a job is finished after the finished file is written
 #: to allow network file system to sync up
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -4,8 +4,6 @@ import hashlib
 import pathlib
 from inspect import isclass, isfunction, ismemberdescriptor
 
-from sisyphus.job_path import AbstractPath
-
 
 def md5(obj):
     """
@@ -99,10 +97,7 @@ def get_object_state(obj):
     # https://github.com/python/cpython/issues/125094
     # Thus, only use __getstate__ if it is not the default object.__getstate__.
     elif hasattr(obj, "__getstate__") and obj.__class__.__getstate__ is not getattr(object, "__getstate__", None):
-        if isinstance(obj, AbstractPath):
-            state = obj._get_state_with_creator()
-        else:
-            state = obj.__getstate__()
+        state = obj.__getstate__()
     else:
         state = _getmembers(obj)
         if not state and not hasattr(obj, "__dict__") and not hasattr(obj, "__slots__"):
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -1,3 +1,4 @@
+from typing import Tuple
 import enum
 import hashlib
 import pathlib
@@ -39,12 +40,42 @@ def short_hash(obj, length=12, chars="0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdef
     return "".join(ls)
 
 
+_BasicTypes: Tuple[type, ...] = (int, float, bool, str, complex)
+_BasicSeqTypes: Tuple[type, ...] = (list, tuple)
+_BasicSetTypes: Tuple[type, ...] = (set, frozenset)
+_BasicDictTypes: Tuple[type, ...] = (dict,)
+_BasicTypesCombined: Tuple[type, ...] = _BasicTypes + _BasicSeqTypes + _BasicSetTypes + _BasicDictTypes
+
+
 def get_object_state(obj):
     """
     Export current object status
 
     Comment: Maybe obj.__reduce__() is a better idea? is it stable for hashing?
     """
+
+    # Note: sis_hash_helper does not call get_object_state in these cases.
+    # However, other code might (e.g. extract_paths),
+    # so we keep consistent to the behavior of sis_hash_helper.
+    if obj is None:
+        return None
+    if isinstance(obj, _BasicTypesCombined):
+        for type_ in _BasicTypesCombined:
+            if isinstance(obj, type_):
+                # Note: For compatibility with old behavior, first only allow if the type is not derived
+                # (thus type(obj) is type_ check).
+                # For derived cases, we need to be more careful.
+                if type(obj) is type_:
+                    return obj
+                if hasattr(obj, "__getnewargs_ex__") or hasattr(obj, "__getnewargs__"):
+                    # E.g. a namedtuple. This is sometimes used, and we must keep the behavior for compatibility.
+                    break  # Use the original behavior.
+                # For now, let's fail, and extend this logic maybe later, to make sure we don't miss anything.
+                # E.g. a dict/tuple/list would contain the elements itself (which should be part of the state),
+                # and maybe *additionally* some other things in __dict__.
+                raise TypeError(f"derived type {obj!r} {type(obj)!r} not handled yet")
+    if isfunction(obj) or isclass(obj):
+        return obj.__module__, obj.__qualname__
 
     if isinstance(obj, pathlib.PurePath):
         # pathlib paths have a somewhat technical internal state
@@ -106,13 +137,17 @@ def sis_hash_helper(obj):
         byte_list.append(obj)
     elif obj is None:
         pass
-    elif type(obj) in (int, float, bool, str, complex):
+    # Note: Using type(obj) in _Types instead of isinstance(obj, _Types)
+    # because of historical reasons (and we cannot change this now).
+    # For derived types (e.g. namedtuple, np.float), it is then handled by get_object_state.
+    # That's why the handling of get_object_state for those types is important.
+    elif type(obj) in _BasicTypes:
         byte_list.append(repr(obj).encode())
-    elif type(obj) in (list, tuple):
+    elif type(obj) in _BasicSeqTypes:
         byte_list += map(sis_hash_helper, obj)
-    elif type(obj) in (set, frozenset):
+    elif type(obj) in _BasicSetTypes:
         byte_list += sorted(map(sis_hash_helper, obj))
-    elif isinstance(obj, dict):
+    elif isinstance(obj, _BasicDictTypes):
         # sort items to ensure they are always in the same order
         byte_list += sorted(map(sis_hash_helper, obj.items()))
     elif isfunction(obj):
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -3,6 +3,8 @@ import hashlib
 import pathlib
 from inspect import isclass, isfunction, ismemberdescriptor
 
+from sisyphus.job_path import AbstractPath
+
 
 def md5(obj):
     """
@@ -66,7 +68,10 @@ def get_object_state(obj):
     # https://github.com/python/cpython/issues/125094
     # Thus, only use __getstate__ if it is not the default object.__getstate__.
     elif hasattr(obj, "__getstate__") and obj.__class__.__getstate__ is not getattr(object, "__getstate__", None):
-        state = obj.__getstate__()
+        if isinstance(obj, AbstractPath):
+            state = obj._get_state_with_creator()
+        else:
+            state = obj.__getstate__()
     else:
         state = _getmembers(obj)
         if not state and not hasattr(obj, "__dict__") and not hasattr(obj, "__slots__"):
--- a/sisyphus/hash.py
+++ b/sisyphus/hash.py
@@ -1,6 +1,7 @@
 import enum
 import hashlib
-from inspect import isclass, isfunction
+import pathlib
+from inspect import isclass, isfunction, ismemberdescriptor
 
 
 def md5(obj):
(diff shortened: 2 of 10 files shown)


critias:
Why not call it `__sis_state__` instead of `_get_state_with_creator` and make use of the already existing check? Thinking about that, `__sis_state__` should have been called something like `__sis_hash_state__`...

It use to be implemented in a similar way before we switched to pickle. At the time the pickled files were only a few MB, so we kept everything inside the `job.save`. That way you can just unpickle the job and look throw the whole tree if you wanted to debug something, but with 100MB files it's not worth it.

I would still recommend to make it an option so it can be turned back on in case it's needed for debugging.

sisyphus/hash.py
```diff
@@ -4,6 +4,8 @@
 import pathlib
 from inspect import isclass, isfunction, ismemberdescriptor
 
+from sisyphus.job_path import AbstractPath
```
michelwi, line 7: Thank you for taking the time of profiling this and additionally for taking the time to contribute the changes to our repo!

I noticed that our tests are failing. `black` can be fixed trivially later, but I think the CI/test might indicate a problem.
We have a .. unorthodox way of importing the `global_settings` so that they can be overridden by user settings in a separate file. This leads to problems with circular imports: in `hash` we now added an import from `job_path` which in turn imports `global_settings` which imports `hash`. And the symptoms can be found in the CI pipeline.
Maybe @critias or @JackTemaki can explain the problem better 🙈 

Maybe moving the import into the `get_object_state` function would already solve it by breaking the loop?

critias:
You are right, that's a circular import. Your suggestion of moving the import should work, but I would prefer the change I suggested above to just renaming `_get_state_with_creator ` to `__sis_state__` and revert all changes in `sisyphus/hash.py`.

RaphaelLilt:
Thanks @critias and @michelwi, I'll incorporate your suggestions and fix the CI tests.

sisyphus/global_settings.py
```diff
@@ -205,6 +205,9 @@ def file_caching(path):
 #: Show job targets on status screen, can significantly slow down startup time if many outputs are used
 SHOW_JOB_TARGETS = True
 
+# Controls whether to include 'creator' in serialized state of an abstract path
+INCLUDE_CREATOR_STATE = True
```
michelwi, line 209: I think we could even default to False and leave the option to be enabled for debugging..?

michelwi (approved):

sisyphus/global_settings.py
critias: I agree, I don't see how enabling it by default would cause any problems and if it does we can simply set it to True again.

RaphaelLilt: Makes sense, I changed it.

RaphaelLilt:
@michelwi I fixed the unit test. For that I removed the `deepcopy` in `sisyphus.job.JobSingleton` which seems to be unnecessary for me and would give an additional speedup. It would be great if you could have a look and verify that this changes does not break anything.

michelwi:
There is some issue with the hashes now. In one example that I looked into there is a `i6_core.returnn.compile.CompileTFGraphJob`  that takes as input a `i6_core.returnn.config.ReturnnConfig` object. In the config I define the dataset as a nested dict that contains `tk.Path` objects which are outputs of `i6_core.returnn.hdf.ReturnnDumpHDFJob`s.
(all referring to https://github.com/rwth-i6/i6_core)

In the current master branch, I have (output of the sis console)
```
In [1]: js = tk.find_job("Compile")
Jobs found:
0: Job<work/i6_core/returnn/compile/CompileTFGraphJob.yK7hwaw8zNIC>

In [2]: p = js[0].__dict__["returnn_config"].config["train"]["files"][0]

In [3]: print(f"{p.creator=} {p.path=}")
p.creator=Job<work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84> p.path='data.hdf'

In [4]: j = tk.find_job("ReturnnDumpHDFJob.6SGIal2fRE84")[0] 
Jobs found:
0: Job<work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84>

In [5]: j.out_hdf.creator
Out[5]: Job<work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84>
```
but with this update there is
```
In [1]: js = tk.find_job("Compile")
Jobs found:
0: Job<work/i6_core/returnn/compile/CompileTFGraphJob.7mZhYhttecOr>

In [2]: p = js[0].__dict__["returnn_config"].config["train"]["files"][0]

In [3]: print(f"{p.creator=} {p.path=}")
p.creator=None p.path='/home/wmichel/code/work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84/output/data.hdf'

In [4]: j = tk.find_job("ReturnnDumpHDFJob.6SGIal2fRE84")[0] 
Jobs found:
0: Job<work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84>

In [5]: j.out_hdf.creator
Out[5]: Job<work/i6_core/returnn/hdf/ReturnnDumpHDFJob.6SGIal2fRE84>
```

Here the hash of the `CompileTFGraphJob` changed and the path that it holds has its creator removed. But the output of the `ReturnnDumpHDFJob` has the creator set correctly in both cases.

michelwi:
The issue seems to be that we like to do deepcopies of our config objects if they are given to multiple jobs and that deepcopy uses `__getstate__` to create the copy, which now breaks the dependence between the jobs.

So, @critias, I am afraid we cannot use your elegant solution, or we need to additionally implement a `__deepcopy__` method that uses `__sis_state__` instead of `__getstate__`.

michelwi (changes requested):
request changes as current state breaks existing setups

michelwi:
Btw. @RaphaelLilt, I assume you are also using sisyphus as your workflow manager. So if you also want to be included in design decisions or checks if changes break existing setups, then feel free to raise a hand or even submit some test cases.

RaphaelLilt:
Thanks, @michelwi, for the comments. Could you please share a minimal example showing that the hashes change between the two versions? I can't reproduce that.
What's the reasoning behind passing a deepcopy of the arguments to the `_sis_init` method? Shouldn't that result in performance issues, similar to the problems that arise when picking up jobs for large graphs this PR intends to fix?

michelwi:
> please share a minimal example

certainly.

```
import copy
from sisyphus import Job, tk

class WriteJob(Job):
    def __init__(self, some_val):
        self.some_val = some_val
        self.out = self.output_path("out.txt")

    def run():
        with open(self.out, 'wt') as outpath:
            outpath.write(self.some_val)

class ReadJob(Job):
    def __init__(self, f):
        self.f = f
        self.out = self.output_path("out.txt")

    def run():
        with open(self.f) as infie:
            for line in infile:
                print(line)
        with open(self.out, "wt") as outfile:
            outfile.write("done")


def py():
    f = WriteJob("Hello World").out
    o = ReadJob(f).out
    tk.register_output("output", o)
    o2 = ReadJob(copy.deepcopy(f)).out
    tk.register_output("output2", o2)
```

expected behavior:
```
$ sis c config/test_config.py

In [1]: tk.find_job("Job")
Jobs found:
0: Job<work/config/test_config/WriteJob.mlrz8RnG9o9U>
1: Job<work/config/test_config/ReadJob.PrhZGzhQOgKP>

$ sis m config/test_config.py
[...]
[2024-10-25 14:53:45,063] INFO: Experiment directory: /home/wmichel/code      Call: /home/wmichel/code/sisyphus/sis m config/test_config.py
[2024-10-25 14:53:45,063] INFO: runnable: Job<work/config/test_config/WriteJob.mlrz8RnG9o9U> <target: output>
[2024-10-25 14:53:45,063] INFO: runnable(1) waiting(1)
```

observed behavior
```
$ sis c config/test_config.py

In [1]: tk.find_job("Job")
Jobs found:
0: Job<work/config/test_config/ReadJob.PrhZGzhQOgKP>
1: Job<work/config/test_config/WriteJob.mlrz8RnG9o9U>
2: Job<work/config/test_config/ReadJob.YSB3u3nSRZJN>

$ sis m config/test_config.py
[...]
[2024-10-25 14:53:28,383] INFO: Experiment directory: /home/wmichel/code      Call: /home/wmichel/code/sis_lillt/sis m config/test_config.py
[2024-10-25 14:53:28,383] ERROR: input_missing: /home/wmichel/code/work/config/test_config/WriteJob.mlrz8RnG9o9U/output/out.txt
[2024-10-25 14:53:28,384] INFO: runnable: Job<work/config/test_config/WriteJob.mlrz8RnG9o9U> <target: output>
[2024-10-25 14:53:28,384] INFO: input_missing(1) runnable(1) waiting(2)
```

The hash of the Job without deepcopy is the same, but in the second case there is one additional Job with a different hash.
Also the manager complains about a missing input, because the deepcopy breaks the dependence between the jobs and sisyphus now assumes that `WriteJob.mlrz8RnG9o9U/output/out.txt` must be an external input.


> What's the reasoning behind passing a deepcopy of the arguments to the _sis_init method?

I was not referring to this particular deepcopy, but rather to generally using deepcopy in my recipes. Imagine that I have prepared my NN training config object (e.g. some very deep nested dict) and now want to do a parameter sweep in training, e.g.
```
config_template = get_it_somehow()
for num_layers in [10, 12, 14, 16]:
    my_config = copy.deepcopy(config_template)
    my_config["params"]["network_config"]["num_layers"] = num_layers
    j = TrainingJob(my_config)
```
then I would employ deepcopy to multiply my config and to avoid side effect interactions between my experiments.

RaphaelLilt:
@michelwi I just pushed a change that should fix your deepcopy issues. Could you please have a look?

michelwi:
I was just copy pasting the next error message:

```
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/local/lib/python3.10/copy.py", line 153, in deepcopy
    y = copier(memo)
  File "/opt/atlassian/pipelines/agent/build/.nox/check_hashes-3-10/lib/python3.10/site-packages/sisyphus/job_path.py", line 252, in __deepcopy__
    return self.__class__(**state)
TypeError: Variable.__init__() got an unexpected keyword argument 'cached'
```
But I saw you updated again - will test that now.

critias:
LGTM, but I leave the final approval to @michelwi since he has been testing it.

michelwi (approved):
I am not so deep into python to know if the deepcopying is done correctly, but all tests pass and an LLM thought the code was good.

RaphaelLilt:
Thanks @critias and @michelwi for the reviews. How do we merge the PR to master?

albertz:
Just as a comment, I wonder if this change could be problematic for other cases. Now here we have the assumption that whenever some Path is pickled/deepcopied, we never want to know about the creator. This might be true for the example given here (and even fix the issues of having too large job.save files). But this assumption is probably not true in general. In general, when you pickle/deepcopy something, you would expect that it contains everything.

It could maybe be solved by having some context manager or so which controls the pickle/deepcopy logic, so when we pickle for job.save, we enable the context manager, which removes some parts for the copy, but in general, when it is not enabled, it would follow the normal expected behavior that everything is copied.

But maybe we can just leave it like it is now until we really get to this problem. However, it might be hard to find the problem later if sth goes wrong because of this.

michelwi:
> whenever some Path is pickled/deepcopied

Your comment is valid for pickling Jobs.
For deepcopy we indeed already ran into a problem and have a separate implementation now that keeps the creator.

RaphaelLilt:
> Your comment is valid for pickling Jobs.

That's right.

>  In general, when you pickle/deepcopy something, you would expect that it contains everything.

I think this flexibility comes with a huge performance cost which makes running larger graphs with >1M jobs extremely slow. We currently trying to make `remove_job_and_descendants` faster, since it takes multiple days.

albertz:
Yes, I'm not against the solution here, I was just saying, this solution is now specific to this problem (and a very reasonable solution for it), but maybe it breaks now in other cases, as this behavior now is a bit unexpected. But I don't think we need to change anything for now.

In general, it's great that you tackle this problem on working with large graphs. I was often frustrated by the speed, and also have done some other optimizations already in the past.

patrick-wilken:
@Zettelkasten noticed these changes break the `TASK_INPUTS_MUST_BE_AVAILABLE = False` case, which was the default/only behavior until recently.

The worker calls `Path.available()` which has two branches depending on whether the Path has a creator. If yes, it is just checked whether the the creator job has finished. If not, the existence of the file is checked, which does not work for the cases where `TASK_INPUTS_MUST_BE_AVAILABLE` matters (mostly path prefixes).

I don't like that we use these prefixes in our recipes, I think they should have never been supported, and we will get rid of them in the long run. But it shows that what @albertz says is correct, by default you would expect to have the complete Path object available in the worker and not having that is error prone.

patrick-wilken:
When I said "supported" I assumed `Job.path_available()`was implemented with these prefixes in mind. But @michelwi said checking whether the Job is finished is probably a shortcut to avoid file system accesses. This would make sense, and in this case you have every right to break path prefixes that happened to work just because of that. 😄

RaphaelLilt:
I'm not sure if I can follow. The only place where `TASK_INPUTS_MUST_BE_AVAILABLE` appears in the repo is [here](https://github.com/rwth-i6/sisyphus/blob/d256f43b69c0dd668b52822ec0016cb8479e5efd/sisyphus/task.py#L153-L158). I don’t see where the worker calling `Path.available()` and whether the `Path` has a creator. From what I see, it’s just checking the file system directly. Could you clarify?

patrick-wilken:
[This](https://github.com/rwth-i6/sisyphus/blob/d256f43b69c0dd668b52822ec0016cb8479e5efd/sisyphus/worker.py#L213) `job._sis_runnable()` check leads to a `Path.available()` call in `Job._sis_all_path_available()`.

So it's not directly in the `TASK_INPUTS_MUST_BE_AVAILABLE` related code, but it's logically related. If `TASK_INPUTS_MUST_BE_AVAILABLE` is false, paths are allowed to be file name prefixes and that's exactly the case where the result of `Path.available()` changes depending on whether the creator is kept or removed, because the check changes from "creator job is finished" to "file exists" and only then those can be different.

(`TASK_INPUTS_MUST_EXIST` would be a more exact name.)

PR #212, opened 2024-10-24
Adapt SLURM engine for multi-node jobs

NeoLegends:
- Wrap calls into `srun` (better resource tracking according to internet, also required to start `--ntasks > 1` at all)
- Add support for log file linking when multiple SLURM tasks are involved

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -413,26 +413,49 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         if os.path.isfile(logpath):
             os.unlink(logpath)
 
-        job_ids = (os.getenv(name, None) for name in ["SLURM_JOB_ID", "SLURM_JOBID", "SLURM_ARRAY_JOB_ID"])
-        engine_logpath = (
-            os.path.dirname(logpath)
-            + "/engine/"
-            + os.getenv("SLURM_JOB_NAME")
-            + "."
-            + next(filter(None, job_ids), "0")
-            + "."
-            + str(slurm_task_id)
-            + "."
-            + os.getenv("SLURM_ARRAY_TASK_ID", "1")
+        job_id = next(
+            filter(None, (os.getenv(name, None) for name in ["SLURM_JOB_ID", "SLURM_JOBID", "SLURM_ARRAY_JOB_ID"])), "0"
         )
-        try:
-            if os.path.isfile(engine_logpath):
+        has_linked_logfile = False
+        engine_logpath_candidates = [
+            (
+                os.path.dirname(logpath)
+                + "/engine/"
+                + os.getenv("SLURM_JOB_NAME")
+                + "."
+                + job_id
+                + "."
+                + str(slurm_task_id)
+                + "."
+                + os.getenv("SLURM_ARRAY_TASK_ID", "1")
+            ),
+            (
+                os.path.dirname(logpath)
+                + "/engine/"
+                + os.getenv("SLURM_JOB_NAME")
+                + "."
+                + job_id
+                + "."
+                + os.getenv("SLURM_ARRAY_TASK_ID", "1")
+            ),
+        ]
+        for engine_logpath in engine_logpath_candidates:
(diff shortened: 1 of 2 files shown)


sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -232,11 +233,12 @@ def submit_helper(self, call, logpath, rqmt, name, task_name, start_id, end_id,
         :param int step_size:
         """
         name = self.process_task_name(name)
-        sbatch_call = ["sbatch", "-J", name, "-o", logpath + "/%x.%A.%a", "--mail-type=None"]
+        out_log_file = logpath + "/%x.%A.%t.%a"
```
michelwi, line 236: this will break all running jobs when sisyphus is updated?

NeoLegends: Ah, yeah. Probably need a fallback here as well.

albertz: I think I would only include this `%t` if there are multiple tasks.

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -393,28 +396,66 @@ def get_default_rqmt(self, task):
...
+                os.path.dirname(logpath)
+                + "/engine/"
+                + os.getenv("SLURM_JOB_NAME")
+                + "."
+                + job_id
+                + "."
+                + os.getenv("SLURM_ARRAY_TASK_ID", "1")
+            ),
+        ]
+        for engine_logpath in engine_logpath_candidates:
```
michelwi, line 442: I don't think we need this.. we do already know if this is a  multi-SLURM-task job. So just select the correct engine_logpath.

michelwi (approved):

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -232,11 +233,15 @@ def submit_helper(self, call, logpath, rqmt, name, task_name, start_id, end_id,
         :param int step_size:
         """
         name = self.process_task_name(name)
-        sbatch_call = ["sbatch", "-J", name, "-o", logpath + "/%x.%A.%a", "--mail-type=None"]
+        out_log_file = f"{logpath}/%x.%A.%a"
+        if rqmt.get("multi_node_slots", 1) > 1:
+            out_log_file += ".%t"
+        sbatch_call = ["sbatch", "-J", name, "--mail-type=None"]
         sbatch_call += self.options(rqmt)
+        sbatch_call += ["-o", f"{out_log_file}.batch"]
```
albertz, line 241: What is written to this log file? So we have two log files now?

albertz (approved):
See my comments, otherwise ok.

sisyphus/simple_linux_utility_for_resource_management_engine.py
NeoLegends: The log file is mostly empty. E.g.:

```
--------------------- Slurm Task Prolog ------------------------
Job ID: 4078734
Job name: i6_core.returnn.training.ReturnnTrainingJob.6iUi8jNHCHUX.run
Host: g-11
Date: Sat Nov  2 11:09:11 EDT 2024
User: mgunz
Slurm account: science
Slurm partition: gpu-24g
Work dir:
------------------
Node usage:
JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
4078734_1   gpu-24g i6_core.    mgunz  R       0:00      1 g-11
------------------
Show launch script with:
sacct -B -j
------------------
--------------------- Slurm Task Prolog ------------------------
```

I think we should keep it as sisyphus cleans up the engine directory anyway, and if the `srun` call fails for some reason I think the logs of that would end up in the `.batch` log file. Otherwise there's also the option to output to `/dev/null` but then we definitely lose any potentially helpful information when things go wrong.

michelwi: >as sisyphus cleans up the engine directory anyway

does it do that? I thought that the folder is always kept..

Anyway, I agree that the logic is fine

PR #214, opened 2024-10-29
Add threads for alias and output updates on startup

### LINKED ISSUE #213: Creating Alias and Outputs on startup are slow
I am currently running into the issue that my manager startup is slowed down by updating all alias and outputs every time. I am wondering, is this even required for the manager to work properly? I guess this can potentially cause files to be "missing" from output and alias, but behaviour itself should be safe right?

https://github.com/rwth-i6/sisyphus/blob/3181d724a999e6ec032656c2fb3e8ae14ed3eb97/sisyphus/manager.py#L551

For context I am currently looking at 52 secs until config load, 280 secs for alias and 1370 secs for outputs. While this probably is also related to slow fs and creating quite a number of outputs/alias these are hard to fix for me right now. In case the behaviour is not endangered I would create a PR adding a flag to disable the full update on startup. 

My first test shows that this should work, but since I am not that familiar with the manager loop I want to make sure this does not implicitly break anything.

Atticus1806:
Fix #213

### DIFF
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -211,6 +211,7 @@ class Manager(threading.Thread):
         self.ui = ui
         self.interactive = interative
         self.interactive_always_skip = set()
+        self.update_out_sem = threading.Semaphore()
 
         self.stop_if_done = True
         self._stop_loop = False
@@ -436,12 +437,13 @@ class Manager(threading.Thread):
             self.thread_pool.map(f, self.jobs.get(gs.STATE_RUNNABLE, []))
 
     def check_output(self, write_output=False, update_all_outputs=False, force_update=False):
-        targets = self.sis_graph.targets if update_all_outputs else self.sis_graph.active_targets
-        for target in targets:
-            target.update_requirements(write_output=write_output, force=force_update)
-            if target.is_done():
-                target.run_when_done(write_output=write_output)
-                self.sis_graph.remove_from_active_targets(target)
+        with self.update_out_sem:
+            targets = self.sis_graph.targets if update_all_outputs else self.sis_graph.active_targets
+            for target in targets:
+                target.update_requirements(write_output=write_output, force=force_update)
+                if target.is_done():
+                    target.run_when_done(write_output=write_output)
+                    self.sis_graph.remove_from_active_targets(target)
 
     def continue_manager_loop(self):
         # Stop loop flag is set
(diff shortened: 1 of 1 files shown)


albertz:
How thread-safe is Sisyphus? There is no potential issue with that?

Atticus1806:
Since this is just alias and output updates I dont think there is a problem. These ops should be executable any time

albertz:
I'm more thinking about all the internal caching mechanisms of jobs, outputs, paths, the graph, etc. All of that must be thread-safe. (But it might be already, as we also use multi-threading elsewhere. I don't recall the details now.)

Atticus1806:
While my impression is they are, I am passing the question to the experts since I dont know for sure.

critias (commented):
I don't see any problem running `create_aliases` in a separate thread, but check_output might be problematic. If it slows down the startup to much you could consider removing it here. It will be called again inside the manager loop anyway.
That should be tested, but I think that's less dangerous than running it in an extra thread.

Side note: It would be nicer to use the already existing manager thread_pool instead of starting new threads.

Atticus1806:
I updated to use the thread pool. 

>It will be called again inside the manager loop anyway.

While that is true, its called with other parameters right? So a "full" update is only done at the beginning. I dont mind removing it here and then saying "user needs to call update for a full update", but this should be discussed.
Or do we add a flag `Skip_Output_update_on_startup` or smth.?

Atticus1806:
I am currently testing the semaphore version, but what I am wondering about is this line:
https://github.com/rwth-i6/sisyphus/blob/d256f43b69c0dd668b52822ec0016cb8479e5efd/sisyphus/manager.py#L485 
This looks quite redundant too me, since it is called in all relevant paths later aswell. Maybe this can be removed here?

albertz:
What is the status here?

sisyphus/manager.py
```diff
@@ -211,6 +211,7 @@ def __init__(
         self.ui = ui
         self.interactive = interative
         self.interactive_always_skip = set()
+        self.update_out_sem = threading.Semaphore()
```
albertz, line 214: Why `Semaphore` and not `Lock`?

Atticus1806:
@critias do you think this is safe now, as it is?

critias:
I don't see any problem anymore (beside changing the Semaphore into a Look).
Thank you!

PR #219, opened 2024-11-12
Clarify the use of Job.set_rqmt and Job.update_rqmt

### LINKED ISSUE #218: Confusing documentation for Job.set_rqmt and Job.update_rqmt
The documentation of update_rqmt says "Updates the given requirements for this job, values not set in rqmt will not be affected.". So I had assumed that every time I called this function, it would update the rqmts of the task with the values of the new rqmt. However, internally this is not the case. The provided rqmt is stored and used later. Therefore, multiple calls overwrite the previous results.

See here:
```
    def update_rqmt(self, task_name, rqmt):
        """Updates the given requirements for this job, values not set in rqmt will not be affected.

        :param str task_name: Which task will be affected
        :param rqmt: the new requirements
        :return:
        """
        self._sis_task_rqmt_overwrite[task_name] = rqmt.copy(), False
        return self
```
This is then used later in the _sis_tasks method
```
                if task_name in self._sis_task_rqmt_overwrite:
                    rqmt, replace = self._sis_task_rqmt_overwrite[task_name]
                    if replace:
                        task._rqmt = rqmt
                    else:
                        task._rqmt.update(rqmt)
```

I was not the only one confused by this so I believe we should clarify the documentation.

See #219 for a proposed fix.

jairsan:
Fixes #218

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -1114,6 +1114,8 @@ class Job(metaclass=JobSingleton):
     def set_rqmt(self, task_name, rqmt):
         """Overwrites the given requirements for this job
 
+        If this function is called more than once, only the effects of the last call are applied.
+
         :param str task_name: Which task will be affected
         :param rqmt: the new requirements
         :return:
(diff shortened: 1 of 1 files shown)


michelwi:
and additionally even set_ and update_ are mutually overwriting each other.

michelwi (approved):

PR #220, opened 2024-11-13
Add Logging filtering and Alias printing to Cleaning

Atticus1806:
When using the sis cleaner I often run into two problems:
1. I clean >5000 jobs, which causes "List affected directories" to be unreadable
2. The list only gives the hash, which is kind of useless when determining if the job is really to be cleaned (since I use subgraphs I want to ensure that the main graph is complete and I dont accidentality clean wrong things)

This PR proposed solutions to both problems. 
The first is introducing the option to give substrings to the clean methods, such that only jobs which fit the substring (e.g. ReturnnTraining or AdvancedTree) will be printed when listing.
The second is reading the info file of the cleaned jobs and trying to get the alias during creation. This of course is purely heuristic and there might be a cleaner solution, but I could not find one, since the alias itself does not exist if the job is not in the graph. 

If there are better solutions to one of the problems feel free to suggest, so that I can update them. 
Also I added some typing/doc.

### DIFF
--- a/sisyphus/cleaner.py
+++ b/sisyphus/cleaner.py
@@ -20,6 +20,8 @@ import shutil
 import sys
 import tempfile
 
+from typing import List, Optional
+
 from sisyphus import graph
 import sisyphus.global_settings as gs
 
@@ -206,7 +208,7 @@ def search_for_unused(job_dirs, current=gs.WORK_DIR, verbose=True):
     return unused
 
 
-def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", force=False):
+def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", force=False, filter_affected=None):
     """list all directories that will be deleted and add a security check"""
     if isinstance(dirs, str):
         dirs = load_remove_list(dirs)
@@ -228,7 +230,19 @@ def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", fo
         if input_var.lower() != "n":
             logging.info("Affected directories:")
             for i in tmp:
-                logging.info(i)
+                if os.path.exists(i + "/info") and gs.CLEANER_PRINT_ALIAS:
+                    with open(i + "/info") as f:
+                        lines = f.readlines()
+                        if lines[-1].strip().startswith("ALIAS"):
+                            s = lines[-1].strip()
+                            s.replace("ALIAS:", "ALIAS AT CREATION:")
+                        else:
+                            s = ""
+                else:
+                    s = ""
+                if filter_affected is None or any(x in i for x in filter_affected):
+                    logging.info(i + "  " + s)
+
     else:
         with tempfile.NamedTemporaryFile(mode="w") as tmp_file:
             for directory in dirs:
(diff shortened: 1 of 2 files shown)


sisyphus/cleaner.py
```diff
@@ -280,8 +295,17 @@ def cleanup_jobs():
...
-    """Go through all jobs in the current graph to remove all jobs with a lower keep value that the given minimum"""
+def cleanup_keep_value(
+    min_keep_value, load_from: str = "", mode: str = "remove", filter_affected: Optional[List[str]] = None
+):
+    """Go through all jobs in the current graph to remove all jobs with a lower keep value that the given minimum
+
+    :param min_keep_value: Remove jobs with lower keep value than this
+    :param load_from: File name to load list with used directories
+    :param mode: Cleanup mode ('remove', 'move', or 'dryrun')
+    :param filter_affected: Defines what substrings should be printed when listing affected directories
```
michelwi, line 306: wow, I am very surprised by `filter_affected`. Because the thing that is affecting the files/directories in these functions is <being deleted>. So I expected the parameter `filter_affected` to filter the folders that are deleted.

But as the docstring clearly states: the affecting only pertains to the being printed part, the folders would still be deleted.

Is this intended? Ok, by the PR title "Add Logging filtering" I assume yes.
But..
* I don't see why one would want to clean everything but only have a specific thing printed to log.
* I do see how one would want to clean up only specific kinds of jobs (e.g. clean up recognitions but keep all the trainings even if unused)

Suggestions:
* change the implementation so that `filter_affected` filters what is cleaned OR
* change the name of the parameter to something like filter_printed / logging_filter / ..

sisyphus/cleaner.py
```diff
@@ -228,7 +231,19 @@ def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", fo
         if input_var.lower() != "n":
             logging.info("Affected directories:")
             for i in tmp:
-                logging.info(i)
+                if os.path.exists(i + "/info") and gs.CLEANER_PRINT_ALIAS:
+                    with open(i + "/info") as f:
+                        lines = f.readlines()
+                        if lines[-1].strip().startswith("ALIAS"):
```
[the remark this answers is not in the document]
Atticus1806: I think there could, I (for myself) decided one would be sufficient, could also try to print all of them if desired.

sisyphus/cleaner.py
Atticus1806: So for me just printing was desired, since I usually have like 10k jobs that want to be cleaned and I just want to print the „longer“ jobs. 
But I can change the name.

sisyphus/cleaner.py
```diff
@@ -241,13 +242,15 @@ def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", fo
                             s = ""
                 else:
                     s = ""
-                if filter_affected is None or any(x in i for x in filter_affected):
+                if filter_printed is None or any(x in i for x in filter_printed) and (filter_affected is None or any(x in i for x in filter_affected)):
```
michelwi, line 245:
```suggestion
                if (filter_printed is None or any(x in i for x in filter_printed)) and (filter_affected is None or any(x in i for x in filter_affected)):
```
missing parentheses

sisyphus/cleaner.py
```diff
@@ -241,13 +242,15 @@ def remove_directories(dirs, message, move_postfix=".cleanup", mode="remove", fo
...
                     logging.info(i + "  " + s)
 
     else:
         with tempfile.NamedTemporaryFile(mode="w") as tmp_file:
             for directory in dirs:
-                tmp_file.write(directory + "\x00")
+                if filter_affected is None or any(x in directory for x in filter_affected):
+                    tmp_file.write(directory + "\x00")
+            time.sleep(1000)
```
michelwi, line 253: why wait so long?

michelwi (changes requested):

sisyphus/cleaner.py
```diff
@@ -80,11 +85,17 @@ def extract_keep_values_from_graph():
     return job_dirs
 
 
-def find_too_low_keep_value(job_dirs, min_keep_value):
+def find_too_low_keep_value(
+    job_dirs: Union[str, Dict[Union[str, Path], int]],
+    min_keep_value: int,
+    filter_removed_jobs: Optional[List[Union[str, Path]]] = None,
+):
```
michelwi, line 92: now that you add typing, please also add the return type..

Atticus1806: that indeed is a good idea...

PR #223, opened 2024-11-27
skip _sis_runnable check when already finished

albertz:
Via: https://github.com/rwth-i6/sisyphus/issues/167#issuecomment-2503658756

See the corresponding discussion there for why I did this.

However, I'm not really sure that this is correct or wanted. I think this mostly makes a potential difference for jobs which have a custom `update` function and which might dynamically add some further dependencies (and maybe also dynamically register further outputs). Now, when such a job is finished, it might never call the `update`, and then you will never get the full graph again. This is what I also want here, to have it faster, but there might be other cases where you would not want this?

I'm also not totally sure about the `return_when_finished` default. But if this default would be None, then there are likely some places where this function is called with this default, and then it anyway would do the update and expand the graph, so that would not really work.

I'm mostly leaving this PR open here for discussion.


michelwi:
> potential difference for jobs which have a custom update function and which might dynamically add some further dependencies

does the MT pipeline contain such things?

sisyphus/job.py
```diff
@@ -684,8 +684,15 @@ def _sis_all_path_available(self):
                 return False
         return True
 
-    def _sis_runnable(self):
+    def _sis_runnable(self, *, return_when_finished: Optional[bool] = True):
         """True if all inputs are available, also checks if new inputs are requested"""
+        if return_when_finished is not None:
```
Atticus1806, line 689: Shouldn't this be `if return_when_finishes is True`? Otherwise putting False would also move into this branch, which would be unexpected to me.

albertz: If you want to do a boolean check, you would just do `if return_when_finishes: ...` (never do `... is True`, that is not standard Python code style).

But the type here intentionally is not just `bool` but `Optional[bool]`. The meaning of `None` is that it should not use it. If it is not None, it should return it. Both False/True can be valid return values.

I could maybe add some docstring for this, like this?
```
:param return_when_finishes: when set, it immediately returns this value when it is finished,
    instead of checking the inputs
```
(Actually, the name maybe should be better `return_when_finished`.)

Note, I made this a possibility to return both True and False for the case it is finished, because it's not clear to me that in all possible cases always returning True is correct, because when this option is used, we don't actually check whether the job is really runnable.

Atticus1806: >never do ... is True, that is not standard Python code style

I know, but I like not giving the use the option to cause weird behaviour by passing something wrong (like `return_when_finished=5`).

But I see your point, I missed that the value itself is returned, then I agree with your variant.

Atticus1806 (approved):

sisyphus/job.py
albertz: > to cause weird behaviour by passing something wrong

If you want to guard against that, I would just do an explicit type check, like:
```python
assert return_when_finished is None or isinstance(return_when_finished, bool)
```

critias:
An side effect of `_sis_runnable` is that it updates the whole dependency tree. Looking at the code I didn't see a case were this change should be a problem, but this change should be tested a while in practice before submitting it.

albertz:
One side effect I realized later: When some finished job never calls its `update` method, there are potentially some jobs missing in the graph. E.g. when you want to perform cleanup, or when you changed your aliases and you want that it updates them, then it would miss those. This might not be so critical, but this could be unexpected. But this is maybe then also a user choice, like faster startup vs having the complete graph?

Zettelkasten:
> does the MT pipeline contain such things?

As far as I can tell, no, we don't override `update()` anywhere.

PR #228, opened 2024-12-02
Add timer to print update

Atticus1806:
When running experiments in subgraphs I often encounter that only 3-4 trainings and nothing else is running. When only doing recognitions on e.g. the last checkpoint this causes the manager to not print an update for a long time. Since the Training job info also includes the current epoch, its desirable for me to print an update every x seconds.  
Thus I propose this flag that if set does this.


michelwi (approved):
yes, I also want that :D

PR #230, opened 2024-12-03
Slurm, round up num CPUs

### LINKED ISSUE #229: log files are broken
For example:
```
$ ls -la work/i6_experiments/users/zeyer/datasets/utils/extract_seq_list/ExtractSeqListJob.5LJd3Q5k1nEN/
total 189
drwxr-xr-x  6 zeyer assi  12 Dec  2 18:24 .
drwxr-xr-x 11 zeyer assi  11 Dec  2 15:08 ..
drwxr-xr-x  2 zeyer assi  12 Dec  3 05:00 engine
-rw-r--r--  1 zeyer assi 334 Dec  2 23:55 info
drwxr-xr-x  2 zeyer assi   3 Dec  2 18:24 input
-rw-r--r--  1 zeyer assi 900 Dec  2 23:55 job.save
lrwxrwxrwx  1 zeyer assi 112 Dec  2 18:24 log.run.1.0 -> engine/i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589181.1.0
lrwxrwxrwx  1 zeyer assi 112 Dec  2 18:24 log.run.1.1 -> engine/i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589181.1.1
drwxr-xr-x  2 zeyer assi   3 Dec  2 18:28 output
-rw-r--r--  1 zeyer assi 905 Dec  3 05:00 submit_log.run
-rw-r--r--  1 zeyer assi 600 Dec  2 18:50 usage.run.1
drwxr-xr-x  2 zeyer assi   2 Dec  2 18:24 work
```
However, in `engine`:
```
$ ls -la work/i6_experiments/users/zeyer/datasets/utils/extract_seq_list/ExtractSeqListJob.5LJd3Q5k1nEN/engine/
total 170
drwxr-xr-x 2 zeyer assi    12 Dec  3 05:00 .
drwxr-xr-x 6 zeyer assi    12 Dec  2 18:24 ..
-rw-r--r-- 1 zeyer assi  8778 Dec  2 18:33 i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589181.1
-rw-r--r-- 1 zeyer assi  2173 Dec  2 18:38 i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589181.1.batch
-rw-r--r-- 1 zeyer assi 10404 Dec  2 18:50 i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589250.1
-rw-r--r-- 1 zeyer assi  2702 Dec  2 18:54 i6_experiments.users.zeyer.datasets.utils.extract_seq_list.ExtractSeqListJob.5LJd3Q5k1nEN.run.9589250.1.batch
```

The job is defined as follows:
```python
class ExtractSeqListJob(Job):
    """
    Takes any dataset dict, and extracts all seq tags from it.
    """

    def __init__(
        self,
        *,
        returnn_dataset: Dict[str, Any],  # to get all seq tags
        returnn_root: Optional[tk.Path] = None,
    ):
        self.returnn_dataset = returnn_dataset
        self.returnn_root = returnn_root

        self.out_seq_list = self.output_path("out_seq_list.txt")

        self.rqmt = {"cpu": 1, "mem": 4, "time": 1, "gpu": 0}

    def tasks(self):
        yield Task("run", rqmt=self.rqmt)

    def run(self):
        ...
```

Multiple things are messed up:

* Why are there `log.run.1.0` and `log.run.0.0`? This is not a batched/multi-node job. Maybe #212 messed this up?
* Those `log.run.*` files are not regular files but symlinks?
* Those `log.run.*` symlinks are broken?
* Those `log.run.*` symlinks are not updated to the latest Slurm job id?

albertz:
Fix #229.

Example: You have `rqmt["cpu"]==1`. Without this change, you get `SLURM_CPUS_PER_TASK=1`. You also by default have `SLURM_JOB_NUM_NODES=1`. Due to the Slurm hyper-threading logic, it might round-up the num CPUs, i.e. you get `NumCPUs=2`. This results in `SLURM_TASKS_PER_NODE=2`.

Without the additional `srun` that was introduced in #212, Slurm will not really handle the `SLURM_TASKS_PER_NODE`, and this problem was not noticed. However, with the `srun` (which should always be fine and follows standard Slurm practice), this is a problem now.

This PR fixes this by avoiding that `SLURM_CPUS_PER_TASK > NumCPUs`.

Test this yourself: Create `slurm-test-script.sh` with content (example taken from the `sbatch` man page):
```
#!/bin/sh
#SBATCH --time=1
srun hostname |sort
```
Run `sbatch --cpus-per-task=1 slurm-test-script.sh`. Then check the output log file.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -163,7 +163,14 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
             gres += str(rqmt["gpu"])
             out.append(gres)
 
-        out.append("--cpus-per-task=%s" % rqmt.get("cpu", 1))
+        def _round_up_to_even(n):
+            return n if n % 2 == 0 else n + 1
+
+        # Use an even number of CPUs, because when we have SLURM_CPUS_PER_TASK=1 (or any uneven num),
+        # but Slurm rounds the NumCPUs up to account for hyper-threading (NumCPUs=2),
+        # this will lead to SLURM_NTASKS=2, i.e. then this job would be executed twice.
+        # https://github.com/rwth-i6/sisyphus/issues/229
+        out.append("--cpus-per-task=%s" % _round_up_to_even(rqmt.get("cpu", 1)))
 
         # Try to convert time to float, calculate minutes from it
         # and convert it back to an rounded string


michelwi:
This is on the i6 cluster.. but will it break things on the apptek cluster? Or other clusters where sisyphus might be used?
@curufinwe do we use hyper-threading?

albertz:
> This is on the i6 cluster.. but will it break things on the apptek cluster? Or other clusters where sisyphus might be used? @curufinwe do we use hyper-threading?

I don't see how this can break anything? In the worst case, you allocate one more CPU than what would be necessary?

NeoLegends (approved):
Agree with Albert, even if we don‘t hyperthrrad (which we likely do?) CPUs don‘t cost much fairshare normally.

michelwi:
> I don't see how this can break anything?

If we have a CPU cluster without hyper-threading and mostly use single-cpu tasks, then this updates would reduce the number of Jobs that can run in parallel by a factor of 2, would it not?
I would call this "a serious problem" if not "broken".

michelwi:
>  ```
> # Use an even number of CPUs, because when we have SLURM_CPUS_PER_TASK=1 (or any uneven num),
> # but Slurm rounds the NumCPUs up to account for hyper-threading (NumCPUs=2),
> # this will lead to SLURM_NTASKS=2, i.e. then this job would be executed twice.
> ```

Just out of curiosity:
what would be the behavior for another uneven number? (e.g. `SLURM_CPUS_PER_TASK=3`)
if rounded up to 4, then both `SLURM_NTASKS=2` (needs 6 cores) and `SLURM_NTASKS=1` (needs 3 cores) is not correct.

albertz:
Hm, I tested this `slurm-test-script.sh` above for other numbers of `--cpus-per-task`, and it seems, for all numbers >=2, I get only one task, it only runs it once? I don't fully understand this yet. This needs some further investigation.

albertz:
I don't really understand it so far. Currently, it seems to me like using `max(..., 2)` instead of `_round_up_to_even(...)` would already work, but I'm not really sure that's correct.

albertz:
As #231 is merged now, this here is obsolete.

PR #231, opened 2024-12-03
Slurm, fix ntasks per node to one

### LINKED ISSUE #229: log files are broken

albertz:
Fix #229.

(Makes PR #230 obsolete.)

(This seems to be needed since we use an additional explicit `srun` via PR #212.)

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -176,6 +176,11 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         if rqmt.get("multi_node_slots", 1) > 1:
             out.append("--ntasks=%s" % rqmt["multi_node_slots"])
             out.append("--nodes=%s" % rqmt["multi_node_slots"])
+        else:
+            # With --cpus-per-task=1, it's sometimes possible that we end up with SLURM_NTASKS=2.
+            # This here prevents this.
+            # https://github.com/rwth-i6/sisyphus/issues/229
+            out.append("--ntasks-per-node=1")
 
         sbatch_args = rqmt.get("sbatch_args", [])
         if isinstance(sbatch_args, str):


michelwi (approved):
I like this one much better than #230

albertz:
Note, for PR #230, at least I had some theory why this would be the correct fix, but apparently my theory was wrong.

For this PR here, while I can see that it might fix the issue, and apparently in my tests, it also does fix the issue, I don't fully understand the effect of this, and why exactly it goes wrong without this.

NeoLegends (approved):

albertz:
I just merged this now, as this is really annoying for me otherwise. But maybe someone can at some later point better explain why this is really the right fix. Or maybe we can just make another followup fix in case we figure out a better way to fix it.

PR #236, opened 2025-01-10
sis_setup_directory: fix crash on broken input symlink

patrick-wilken:
Small thing that annoys me from time to time: linking the job inputs fails with "File exists" if the creator path does not exist (and multiple inputs have same creator) because `isdir(broken_symlink) == False`.
If you run into this, clearly something else is wrong, but should fail at another place, not here.


michelwi (approved):
could we even remove the `isdir` check now?

patrick-wilken:
So, should I change to `exists`?

michelwi:
> So, should I change to `exists`?

I have no strong opinion and accept either way.. you may go ahead and implement the more robust check.

patrick-wilken:
Okay, done. Also quickly checked that it works as expected.

michelwi (approved):

PR #237, opened 2025-01-19
Slurm, fix when running sis worker outside Slurm env

albertz:
You might want to run `sis worker ...` directly to debug something.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -426,7 +426,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         engine_logpath = (
             os.path.dirname(logpath)
             + "/engine/"
-            + os.getenv("SLURM_JOB_NAME")
+            + os.getenv("SLURM_JOB_NAME", self.process_task_name(task.task_name()))
             + "."
             + job_id
             + "."


sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -426,7 +426,7 @@ def init_worker(self, task):
         engine_logpath = (
             os.path.dirname(logpath)
             + "/engine/"
-            + os.getenv("SLURM_JOB_NAME")
+            + os.getenv("SLURM_JOB_NAME", self.process_task_name(task.task_name()))
```
michelwi, line 429: this might get the correct path again when the env var is not set, but will the worker write the log into that file if it is not started via slurm? Would it not be printed into stdout/stderr of the terminal?

sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -426,7 +426,7 @@ def init_worker(self, task):
...
             + "."
             + job_id
             + "."
```
michelwi, line 432: would you also update the default of SLURM_ARRAY_TASK_ID to get the actual task number instead of a dummy 1?

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: > Would it not be printed into stdout/stderr of the terminal?

Yes it prints on stdout.

But this code crashes with an exception, because it tries to do `"...some string..." + None`.

sisyphus/simple_linux_utility_for_resource_management_engine.py
albertz: How do I get that?

The `array_task_id` here in this function is also via the env var. Maybe I could use `array_task_id` instead to not have the logic duplicated, but it's just the same.

I think the env var is the only source we can know the array task id here, or not?

michelwi: ok, if the symlink is anyway unused / broken, we don't need to change this.

PR #243, opened 2025-02-28
Sort data written to job info file

NeoLegends:
This has always annoyed me before. This should make it much easier to find job inputs in a large list of them.

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -291,7 +291,7 @@ class Job(metaclass=JobSingleton):
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
             for tag in sorted(self.tags):
                 f.write("TAG: %s\n" % tag)
-            for i in sorted(p.get_path() if isinstance(p, AbstractPath) else str(p) for p in self._sis_inputs):
+            for i in sorted(p.get_path() for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
             for key in sorted(self._sis_kwargs.keys()):
                 value = self._sis_kwargs[key]
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -293,14 +293,13 @@ class Job(metaclass=JobSingleton):
                 f.write("TAG: %s\n" % tag)
             for i in sorted(p.get_path() for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
-            for key in sorted(self._sis_kwargs.keys()):
-                value = self._sis_kwargs[key]
+            for key, value in sorted(self._sis_kwargs.items()):
                 try:
                     f.write("PARAMETER: %s: %s\n" % (key, value))
                 except UnicodeEncodeError as e:
                     f.write("PARAMETER: %s: <UnicodeEncodeError: %s>\n" % (key, e))
             if self._sis_aliases:
-                for alias in sorted(str(a) for a in self._sis_aliases):
+                for alias in sorted(self._sis_aliases):
                     f.write("ALIAS: %s\n" % alias)
             for stacktrace in self._sis_stacktrace:
                 f.write("STACKTRACE:\n")
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -289,7 +289,7 @@ class Job(metaclass=JobSingleton):
             pickle.dump(self, f)
 
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
-            for tag in sorted(str(t) for t in self.tags):
+            for tag in sorted(self.tags):
                 f.write("TAG: %s\n" % tag)
             for i in sorted(str(p) for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -291,7 +291,7 @@ class Job(metaclass=JobSingleton):
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
             for tag in sorted(self.tags):
                 f.write("TAG: %s\n" % tag)
-            for i in sorted(str(p) for p in self._sis_inputs):
+            for i in sorted(p.get_path() if isinstance(p, AbstractPath) else str(p) for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
             for key in sorted(self._sis_kwargs.keys()):
                 value = self._sis_kwargs[key]
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -289,17 +289,18 @@ class Job(metaclass=JobSingleton):
             pickle.dump(self, f)
 
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
-            for tag in self.tags:
+            for tag in sorted(str(t) for t in self.tags):
                 f.write("TAG: %s\n" % tag)
-            for i in self._sis_inputs:
+            for i in sorted(str(p) for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
-            for key, value in self._sis_kwargs.items():
+            for key in sorted(self._sis_kwargs.keys()):
+                value = self._sis_kwargs[key]
                 try:
                     f.write("PARAMETER: %s: %s\n" % (key, value))
                 except UnicodeEncodeError as e:
                     f.write("PARAMETER: %s: <UnicodeEncodeError: %s>\n" % (key, e))
             if self._sis_aliases:
-                for alias in self._sis_aliases:
+                for alias in sorted(str(a) for a in self._sis_aliases):
                     f.write("ALIAS: %s\n" % alias)
             for stacktrace in self._sis_stacktrace:
                 f.write("STACKTRACE:\n")


michelwi (approved):
Also it makes diffing the info files easier to check what changed. lgtm.

Atticus1806 (approved):

sisyphus/job.py
```diff
@@ -289,17 +289,18 @@ def _sis_setup_directory(self, force=False):
...
 
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
-            for tag in self.tags:
+            for tag in sorted(str(t) for t in self.tags):
                 f.write("TAG: %s\n" % tag)
-            for i in self._sis_inputs:
+            for i in sorted(str(p) for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
-            for key, value in self._sis_kwargs.items():
+            for key in sorted(self._sis_kwargs.keys()):
```
albertz, line 296: I would prefer to keep the original order. The order is well defined here (insertion order).

sisyphus/job.py
```diff
@@ -289,17 +289,18 @@ def _sis_setup_directory(self, force=False):
             pickle.dump(self, f)
 
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
-            for tag in self.tags:
+            for tag in sorted(str(t) for t in self.tags):
```
albertz, line 292: Why is the `str` needed?

albertz, line 292: The main problem is that `self.tags` (`self._sis_tags`), `self._sis_inputs` and `self._sis_aliases` are `set`s, and the order of a `set` is non-deterministic and arbitrary. We could also replace `set` by some similar structure which keep a well defined order (insertion order). Or otherwise `sorted` is also fine, definitely better than the current code.

sisyphus/job.py
```diff
@@ -289,17 +289,18 @@ def _sis_setup_directory(self, force=False):
...
                 f.write("TAG: %s\n" % tag)
-            for i in self._sis_inputs:
+            for i in sorted(str(p) for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
```
albertz, line 295: Btw, while this PR changes this code, I wanted to comment (but maybe that should go into a separate PR/issue): This printing of `INPUT` here is a bit annoying for `tk.Variable`, because in `Variable`, we have this:
```python
    def __str__(self):
        return str(self.get())
```
So it will not print the path but instead it will print the content of the `Variable`. So that means that just looking at this `info` file, you cannot easily see the dependencies (inputs) anymore.

I wonder if this was really intended that way or if this is a bug. At least for my typical use cases, I would like to know the actual path instead.

michelwi: If I recall correctly from the top of my head, then for DelayedBase and derived objects the convention was that `__str__` should yield the correctly serialized value to be written in the config and `__repr__` should be save to call in any case (e.g. if variable is unset) and might give such info as the dependencies.

Which is not the case for DelayedBase https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/sisyphus/delayed_ops.py#L88-L89
But for Path https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/sisyphus/job_path.py#L292-L296 and Variable https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/sisyphus/job_path.py#L492-L499 we get such a nice representation.

So maybe we can implement a similar logic also for DelayedBase and friends and then change this line here to
```
for i in sorted(repr(p) if isinstance(p, DelayedBase) else str(p) for p in self._sis_inputs):
```

albertz: We should first discuss what we actually want to have in the `info` file. Currently, it looks like this, here one example:
```
INPUT: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/oggzip/BlissToOggZipJob.RvwLniNrgMit/output/out.ogg.zip
INPUT: /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/oggzip/BlissToOggZipJob.NSdIHfk1iw2M/output/out.ogg.zip
PARAMETER: returnn_dataset: {'class': 'OggZipDataset', 'path': [<Path /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/oggzip/BlissToOggZipJob.RvwLniNrgMit/output/out.ogg.zip>, <Path /u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/oggzip/BlissToOggZipJob.NSdIHfk1iw2M/output/out.ogg.zip>], 'audio': None, 'targets': None, 'fixed_random_subset': 3000, 'fixed_random_seed': 1}
PARAMETER: returnn_dataset_ext_non_hashed: {'use_cache_manager': True}
PARAMETER: returnn_root: None
```
Note, I have some scripts, which simply parse the `info` file, with this simple code, e.g. to iterate through the Sisyphus graph:
```python
for line in open(".../info").read().splitlines():
  if line.startswith("INPUT: "):
    filename = line[len("INPUT: "):]
    ...
```
So this is quite convenient.

Only recently, I noticed that this does not work correctly when there are `Variable`s in the input, because then it does not store the filename but the content. Then it looks like:
```
INPUT: /rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/work/i6_core/recognition/scoring/ScliteJob.dulwk4XbnQMj/output/reports                          
INPUT: 2.12
INPUT: /rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/work/i6_core/recognition/scoring/ScliteJob.vWhUos03wE9T/output/reports                          
INPUT: /rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/work/i6_core/recognition/scoring/ScliteJob.rjgRsNFKyZrQ/output/reports                          
INPUT: 10.23
INPUT: 10.53
...
```

Was this really intentionally to have it like this in the `info` file? I would argue, it is more consistent, makes more sense, easier to parse, more useful, when it just always prints the input filename, always in the simple form:
```
INPUT: <filename>
```
What is the purpose of this `info` file anyway?

Yes, for programming some scripts, to get the Sisyphus graph, I could also load the pickled job object state from `job.save`, or I could run the Sisyphus config. But both are much much slower, and come with further problems.

michelwi: I don't know what the original intention was anyway, but if always having the path works best for you we can have it like that.
```
for i in sorted(p.get_path() if isinstance(p, AbstractPath) else str(p) for p in self._sis_inputs):
```
should do the trick.

Except if your input is a Delayed Function of one or more variables:
```
num_a: tk.Variable = CountSomethingJob(file_a).out_num
num_b: tk.Variable = CountSomethingJob(file_b).out_num
product: delayed_ops.DelayedProduct = num_a * num_b
SomeOtherJob(number=product)
```
now there is no single filesystem path that represents the input

But independent of the desired implementation of your issue, we ca nmerge this PR as is now and come back to proper printing of variables later, right?

albertz: But everything in `_sis_inputs` should always be `AbstractPath`? It all comes `tools.extract_paths`. Well, actually, the `extract_paths` has this check:
```python
        if hasattr(obj, "_sis_path") and obj._sis_path is True and type(obj) is not type:
            out.add(obj)
```
The only two cases which we currently have which would satisfy this are object instances of type `AbstractPath` or `OutputPath`.

albertz: > But independent of the desired implementation of your issue, we can merge this PR as is now and come back to proper printing of variables later, right?

Well, the sorting again would be different then. It's a simple code change, so maybe could be done right here. But I don't mind. It can also be another separate PR.

sisyphus/job.py
NeoLegends: I see value in both options. Personally, I prefer `sorted` because it allows you to binary search w/ your eyes when you are looking for a specific value.

NeoLegends: Ah, they are already strings?

sisyphus/job.py
```diff
@@ -289,17 +289,18 @@ def _sis_setup_directory(self, force=False):
...
-            for key, value in self._sis_kwargs.items():
+            for key in sorted(self._sis_kwargs.keys()):
+                value = self._sis_kwargs[key]
                 try:
                     f.write("PARAMETER: %s: %s\n" % (key, value))
                 except UnicodeEncodeError as e:
                     f.write("PARAMETER: %s: <UnicodeEncodeError: %s>\n" % (key, e))
             if self._sis_aliases:
-                for alias in self._sis_aliases:
+                for alias in sorted(str(a) for a in self._sis_aliases):
```
albertz, line 303:
```suggestion
                for alias in sorted(self._sis_aliases):
```

sisyphus/job.py
```diff
@@ -289,17 +289,18 @@ def _sis_setup_directory(self, force=False):
...
         with open(self._sis_path(gs.JOB_INFO), "w", encoding="utf-8") as f:
-            for tag in self.tags:
+            for tag in sorted(self.tags):
                 f.write("TAG: %s\n" % tag)
-            for i in self._sis_inputs:
+            for i in sorted(p.get_path() for p in self._sis_inputs):
                 f.write("INPUT: %s\n" % i)
-            for key, value in self._sis_kwargs.items():
+            for key in sorted(self._sis_kwargs.keys()):
+                value = self._sis_kwargs[key]
```
albertz, line 297:
```suggestion
            for key, value in sorted(self._sis_kwargs.items()):
```

albertz (approved):

NeoLegends:
This should get the CI fix from #245 before being merged.

PR #247, opened 2025-03-17
Declare `gs.active_engine`, add engine method to get hostnames of job nodes

NeoLegends:
Declares typings for `gs.active_engine` so that users can find it and adds a function to the engine that returns the list of host names/nodes the current job is executing on.

I want to use this for multi-node trainings to determine a master node that coordinates node rendezvous. I can either do this statically by pre-determining in a very exact way which nodes the job is going to execute on (not so flexible), or dynamically by querying SLURM env vars (much more flexible).

I am also trying out doing the same thing via MPI (via mpi4py), which is why this is only a draft. Probably the route via MPI is better, more complete and more generic and does not reinvent the wheel, so I might end up closing this PR again. This route, however, does not depend on MPI, so maybe there is also value in that (no need for the mpi4py python lib).

EDIT: I haven't been successful with MPI so far.


sisyphus/simple_linux_utility_for_resource_management_engine.py
```diff
@@ -456,3 +456,18 @@ def process_task_name(self, name):
             name = self.job_name_mapping(name)
         name = name.replace("/", ".")  # escape name
         return name
+
+    def get_job_node_hostnames(self):
```
albertz, line 460: Some docstring should be added here. Just list some example output.

sisyphus/global_settings.py
```diff
@@ -6,14 +6,22 @@
...
     from sisyphus import Job
+    from sisyphus.engine import EngineBase
+
+
+"""
+In a worker context, exposes the engine the current job is executing on.
+
+In the manager context, this value is `None`.
+"""
+active_engine: Optional[EngineBase] = None
```
michelwi, line 24: what would this do? It does not seem like a global setting to me that a user should update and it seems to never be changed in sisyphus?

sisyphus/engine.py
```diff
@@ -201,6 +202,18 @@ def init_worker(self, task):
         """
         pass
 
+    def get_job_node_hostnames(self) -> List[str]:
```
michelwi, line 205: would you implement this function also in `EngineSelector` and the other engines we have implemented?

sisyphus/localengine.py
```diff
@@ -341,3 +342,9 @@ def get_task_id(self, task_id):
             "Job in local engine started without task_id, worker is probably started manualy. Continue with task_id=1"
         )
         return 1
+
+    def get_job_node_hostnames(self) -> List[str]:
+        """
+        :return: we are running on `["localhost"]` since this is a local engine.
+        """
+        return ["localhost"]
```
michelwi, line 350: is the string "localhost" enough for your purposes or should we try to determine the hostname of the local host?

michelwi (approved):

albertz (approved):

PR #251, opened 2025-03-25
Fix job IDs in SLURM log symlink not matching up

NeoLegends:
It turns out, %A refers to different things in `sbatch` and in the wrapped `srun`, leading to broken log symlinks in the job folder :( 

%j seems to be consistent: https://slurm.schedmd.com/sbatch.html, I also confirmed locally.

Current behavior (pre-fix):
```
❯ ls work/config/create_corpora/CreateCsvFromAvroJob.Z8lxU0d8s4Dt/engine
config.create_corpora.CreateCsvFromAvroJob.Z8lxU0d8s4Dt.run.8344989.1
```

```
❯ ll work/config/create_corpora/CreateCsvFromAvroJob.Z8lxU0d8s4Dt/log.run.1
lrwxrwxrwx 1 mgunz domain_users 76 Mar 25 07:54 work/config/create_corpora/CreateCsvFromAvroJob.Z8lxU0d8s4Dt/log.run.1 -> engine/config.create_corpora.CreateCsvFromAvroJob.Z8lxU0d8s4Dt.run.8345118.1
```

```
❯ scontrol show jobid 8345118
JobId=8345118 ArrayJobId=8344989 ArrayTaskId=1 JobName=config.create_corpora.CreateCsvFromAvroJob.Z8lxU0d8s4Dt.run
   [...]
   StdOut=/nas/data/speech/untranscribed/EVDS/setups/2025-03-24_mgunz_corpus/work/config/create_corpora/CreateCsvFromAvroJob.Z8lxU0d8s4Dt/engine/config.create_corpora.CreateCsvFromAvroJob.Z8lxU0d8s4Dt.run.8344989.1.batch
```

With the fix the job directory looks as usual/intended with the log files from the job main directory hardlinked to the log files in the engine directory. We also have .batch log files in the engine directory (since that PR), which contain the logs of the srun invocation when the job has been scheduled, but not any logs of the sisyphus job itself.

### DIFF
--- a/sisyphus/simple_linux_utility_for_resource_management_engine.py
+++ b/sisyphus/simple_linux_utility_for_resource_management_engine.py
@@ -238,7 +238,7 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         :param int step_size:
         """
         name = self.process_task_name(name)
-        out_log_file = f"{logpath}/%x.%j.%a"
+        out_log_file = f"{logpath}/%x.%A.%a"
         if rqmt.get("multi_node_slots", 1) > 1:
             out_log_file += ".%t"
         sbatch_call = ["sbatch", "-J", name, "--mail-type=None"]
@@ -420,13 +420,12 @@ class SimpleLinuxUtilityForResourceManagementEngine(EngineBase):
         if os.path.isfile(logpath):
             os.unlink(logpath)
 
-        job_id = next(filter(None, (os.getenv(name, None) for name in ["SLURM_JOB_ID", "SLURM_JOBID"])), "0")
         engine_logpath = (
             os.path.dirname(logpath)
             + "/engine/"
             + os.getenv("SLURM_JOB_NAME", self.process_task_name(task.task_name()))
             + "."
-            + job_id
+            + os.getenv("SLURM_ARRAY_JOB_ID", "0")
             + "."
             + os.getenv("SLURM_ARRAY_TASK_ID", "1")
         )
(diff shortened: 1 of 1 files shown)


curufinwe:
I don't quite understand the example. Is this after your fixes or before your fixes? Looking at the code you have now two log file locations. One for the sbatch call and one for the srun call. The sbatch call has the .batch suffix. It seems you do not want to use this one. Correct?

Could you give examples of the current behavior and the behavior after this PR is merged?

NeoLegends:
Ah, right. This is before the fix. You can see that sisyphus creates the log files in the engine directory with one ID and then attempts to symlink to a file using a different ID. Now, using %j instead of %A, the file name in the engine directory and the attempted symlink match up like before and the job folder looks as usual.

curufinwe:
OK, so it seems that the srun execution get's a new array-job-id. Do we then want to remove SLURM_ARRAY_JOB_ID from the job_id detection in `init_worker`? Also: did you test jobs that submit real job arrays?

NeoLegends:
>Do we then want to remove SLURM_ARRAY_JOB_ID from the job_id detection in init_worker?

Good point! Done.

> Also: did you test jobs that submit real job arrays?

Yes.

michelwi:
Is the ID in the engine folder now inconsistent with the ID as is printed in squeue and/or the ID in the prologue output.

e.g. if I have a Job with subtasks and some of them failed and got re-submitted in different sbatch calls and I want to identify which `log.run.i` belongs to which line in `squeue` or can be cancelled with which `scancel $i` call.. is that still clear?

NeoLegends:
Argh, no, it does not seem so. Let me check if there is a more reliable way to get the array job master ID from srun.

NeoLegends:
So maybe it was never wrong to use %A, but the init_worker script must not look at SLURM_JOBID but rather at SLURM_ARRAY_JOB_ID only.

NeoLegends:
Yeah, this seems ok now, at least for an array. job

NeoLegends:
@curufinwe Also tested this for a non-array job and it works there.

curufinwe (approved):

PR #252, opened 2025-04-17
Add gs.on_job_finished callback

dorian-K:
I need this to update permissions on certain output files of some jobs. This seems like a good generic solution that can be used in other situations as well, just like `gs.on_job_failure`

### DIFF
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -7,7 +7,7 @@ These settings can be overwritten via a ``settings.py`` file in the current dire
 import logging
 import sys
 from typing import Dict
-
+from future import __annotations__
 import sisyphus.hash
 from sisyphus.global_constants import *
 from typing import TYPE_CHECKING
--- a/sisyphus/global_settings.py
+++ b/sisyphus/global_settings.py
@@ -85,6 +85,19 @@ def on_job_failure(job):
     pass
 
 
+def on_job_finished(job):
+    """
+    Job finished hook.
+
+    Can be used to update permissions on job outputs after completion.
+
+    The callback needs to be stateless and indempotent, as it can be called multiple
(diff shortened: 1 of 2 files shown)


sisyphus/global_settings.py
```diff
@@ -85,6 +85,19 @@ def on_job_failure(job):
     pass
 
 
+def on_job_finished(job):
```
albertz, line 88:
```suggestion
def on_job_finished(job: Job):
```

albertz (changes requested):
See comment.
And check black.

sisyphus/global_settings.py
dorian-K: We can't import Job directly because this would create a circular dependency. I tried putting the import behind TYPE_CHECKING but now the tests are failing :(

dorian-K:
I think the black test is broken? Not sure how to interpret that error

albertz:
You also need:
From future import annotations

sisyphus/global_settings.py
```diff
@@ -10,6 +10,10 @@
```
albertz, line 10:
```suggestion
from future import __annotations__
```

sisyphus/global_settings.py
albertz: See my comment. That should work.

albertz:
I think the CI error on black is unrelated to the change. It seems like some temporary hiccup on some caching mechanism somewhere?

albertz (approved):

NeoLegends (approved):

albertz:
As an alternative solution: I think, adding `umask 002` to the `~/.zshrc` or `~/.bashrc` should also work. umask is set to 007 by default, which results in by-other-non-readable files.

PR #253, opened 2025-04-19
Add support for using `round(x, n)` on delayed items

NeoLegends:
Previously calling `round(sisyphus_var, n)` caused an error.

### DIFF
--- a/sisyphus/delayed_ops.py
+++ b/sisyphus/delayed_ops.py
@@ -1,4 +1,6 @@
-from typing import Any, Iterable
+from __future__ import annotations
+
+from typing import Union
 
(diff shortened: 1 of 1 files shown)


sisyphus/delayed_ops.py
```diff
@@ -61,6 +63,9 @@ def __pow__(self, other, modulo=None):
     def __rpow__(self, other):
         return DelayedPow(other, self)
 
+    def __round__(self, n: Union[int, DelayedBase] = None):
```
albertz, line 66:
```suggestion
    def __round__(self, n: Union[None, int, DelayedBase] = None):
```

albertz (approved):

Issue #195, opened 2024-07-03
Drop older Python support / increase min required Python version


albertz:
I just want to open this issue to discuss this, and also on what version we want to increase to.

On RETURNN, currently we require Python 3.7. See https://github.com/rwth-i6/returnn/issues/487 for some discussion and also what features are available since Python 3.7 (or closeby versions). But there is also another issue to maybe increase this further (https://github.com/rwth-i6/returnn/issues/1326).

Python >=3.8 is now required for i6_core: https://github.com/rwth-i6/i6_core/pull/409, https://github.com/rwth-i6/i6_core/issues/408

Same discussion for Python >=3.8 for i6_models: https://github.com/rwth-i6/i6_models/pull/11

albertz:
Ping? @curufinwe @michelwi @JackTemaki ?

michelwi:
Bumping would be fine with me, as all our recipes already require `Python >=3.8` as you said.

@dthulke, @patrick-wilken @RaphaelLilt you are using different recipe collections, would bumping the minimum required python version break your workflows?

RaphaelLilt:
I'm currently using Python 3.11 so the bump would be fine with me.

patrick-wilken:
Same here, we're on 3.10.

michelwi:
> so I guess it is Python 3.8 then. Or even newer?
I would propose 3.8 for now and we can come back to this once we need newer features.

> Where should this be put? CI run with this, so for the GitHub Actions configs. 

#245 takes care of this 👍 

> Maybe also README somewhere. 

yes, this is written explicitly in the readme https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/README.md?plain=1#L10

> What else?

Here is what I found with a quick search:
https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/tox.ini#L2 
https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/pyproject.toml#L3
https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/.pre-commit-config.yaml#L7
https://github.com/rwth-i6/sisyphus/blob/fabafb8699ef298ac00f237e7e773c1fac216cd5/docs/installation.rst?plain=1#L15
[...]

PR #255, opened 2025-04-23
SISGraph remove_from_active_targets optimized

albertz:
I noticed that during startup, I had a hang for about 3 minutes in:
```
...
Manager.check_output
SISGraph.remove_from_active_targets
OutputTarget.__eq__
```
I guess all the iterations through the dicts were slow when the list of active targets was huge. Also, obviously this was O(N^2) here, as it iterates always through the whole list.

I could make this a set, but then it would lose the ordering (sets are unordered). A dict has very similar properties, but it is ordered.

### DIFF
--- a/sisyphus/graph.py
+++ b/sisyphus/graph.py
@@ -243,7 +243,7 @@ class SISGraph(object):
 
     def __init__(self):
         self._targets = set()  # type: set[OutputTarget]
-        self._active_targets = []  # type: list[OutputTarget]
+        self._active_targets = {}  # type: dict[OutputTarget,None]  # dict to keep it sorted
         self._pool = None
         self.used_output_path = set()
 
@@ -260,10 +260,10 @@ class SISGraph(object):
 
     @property
     def active_targets(self):
-        return self._active_targets
+        return self._active_targets.keys()
 
     def remove_from_active_targets(self, target):
-        self._active_targets = [out for out in self._active_targets if out != target]
+        self._active_targets.pop(target, None)
 
     @property
     def targets_dict(self):
@@ -314,7 +314,7 @@ class SISGraph(object):
             pass
 
         if not target.is_done():
-            self._active_targets.append(target)
+            self._active_targets[target] = None
 
     def update_nodes(self):
         """Update all nodes to get the most current dependency graph"""


NeoLegends (approved):

critias:
> A dict has very similar properties, but it is ordered.

We should bump the required python version to 3.7 for this to be true. On the other hand I don't see any reason why the order matters here, it's only used by the manager [here](https://github.com/rwth-i6/sisyphus/blob/a48d6ee3a8a1cce63db8b4fb2664e8d72ea46143/sisyphus/manager.py#L450 ) and this shouldn't depend on the target order.
So using a set here should be fine.

albertz:
> We should bump the required python version to 3.7 for this to be true

Yes, I don't think anyone uses an older Python anymore. I was kind of expecting that here. (And even if not, as you say, it's not totally critical.) But making this bump official is #195.

> On the other hand I don't see any reason why the order matters here, it's only used by the manager [here](https://github.com/rwth-i6/sisyphus/blob/a48d6ee3a8a1cce63db8b4fb2664e8d72ea46143/sisyphus/manager.py#L450) and this shouldn't depend on the target order.

Yes, it's used there in `Manager.check_output`, specifically this loop:
```python
targets = self.sis_graph.targets if update_all_outputs else self.sis_graph.active_targets
for target in targets:
    target.update_requirements(write_output=write_output, force=force_update)
    if target.is_done():
        target.run_when_done(write_output=write_output)
        self.sis_graph.remove_from_active_targets(target)
```
Actually, I would argue, we should also make the order of `SISGraph.targets` deterministic. The `target.run_when_done` will do some things, e.g. creating symlinks, logging, reporting, etc. The `target.update_requirements` will/can indirectly also call `_sis_finished`, `_sis_setup`, `_sis_runnable`, etc. I just prefer it if the order of handling these is deterministic (by the order those outputs are actually registered).

critias:
Fair point, having it deterministic has it's advantages. I just wanted to point out that it's not needed as far as I can tell.

albertz:
Yes. I added a comment to clarify.

albertz:
Oh, during another test run, now I got this:
```
  File "/rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/tools/sisyphus/sisyphus/tools.py", line 303, in default_handle_exception_interrupt_main_thread.<locals>.wrapped_func
    line: return func(*args, **kwargs)
    locals:
      func = <local> <function Manager.run at 0x1467d27d2fc0>
      args = <local> (<Manager(Thread-21, initial)>,)
      kwargs = <local> {}
  File "/rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/tools/sisyphus/sisyphus/manager.py", line 655, in Manager.run                                        
    line: self.check_output(write_output=self.link_outputs)
    locals:
      self = <local> <Manager(Thread-21, initial)>
      self.check_output = <local> <bound method Manager.check_output of <Manager(Thread-21, initial)>>                                                                   
      write_output = <not found>
      self.link_outputs = <local> True
  File "/rwthfs/rz/cluster/home/az668407/setups/combined/2021-05-31/tools/sisyphus/sisyphus/manager.py", line 453, in Manager.check_output                               
    line: for target in targets:
    locals:
      target = <local> <sisyphus.graph.OutputPath object at 0x14667eba3ec0>
      targets = <local> dict_keys([<sisyphus.graph.OutputPath object at 0x1467cb8f95b0>, <sisyphus.graph.OutputPath object at 0x1467ca8ea990>, <sisyphus.graph.OutputPath object at 0x1467d65c21b0>, <sisyphus.graph.OutputPath object at 0x1467cb9de3f0>, <sisyphus.graph.OutputPath object at 0x1467d65f8980>, <sisyphus.graph...., len = 26861  
RuntimeError: dictionary changed size during iteration
```
This is not due to having a `dict` instead of a `list` now. Before, we did not have this problem because `remove_from_active_targets` did not remove from the list, it created a new list:
```python
def remove_from_active_targets(self, target):
    self._active_targets = [out for out in self._active_targets if out != target]
```
I'm not exactly sure how to handle this...

albertz:
> I'm not exactly sure how to handle this...

Specifically:

* `SISGraph.active_targets`, should this return a copy or a ref? If a ref, like right now, it means I should fix `Manager.check_output`, do to a copy there.
* Should `Manager.check_output` actually also handle newly added targets?

critias:
> * `SISGraph.active_targets`, should this return a copy or a ref? If a ref, like right now, it means I should fix `Manager.check_output`, do to a copy there.

I think `SISGraph.active_targets` should return a copy.

> * Should `Manager.check_output` actually also handle newly added targets?

This would only be relevant if new targets are added while `Manager.check_output` runs. I think these targets can wait till the next loop as long as we don't have a counter example where this is a problem. It wasn't so far.

albertz:
> I think `SISGraph.active_targets` should return a copy.

I noticed, `SISGraph.targets` is also not a copy.

Also, the only other use despite `Manager.check_output` is in `Manager.continue_manager_loop` with `len(self.sis_graph.active_targets) == 0`, where returning a copy would be quite wasted.

So I decided to do a copy in `Manager.check_output` now.

PR #258, opened 2025-05-07
Fix stack depth check

### LINKED ISSUE #259: Increase stack depth intelligibility
With the current repository version, the `gs.JOB_ADD_STACKTRACE_WITH_DEPTH` variable doesn't work with a value of `-1` or `None` as intended.

#258 fixes the issue that comes by setting the variable to `-1`. However, `None` is still not supported, since setting it results in `TypeError: '>=' not supported between instances of 'NoneType' and 'int'`.

This issue tracks changes and issues in the variable that represents the stack depth.

Icemole:
The stack depth check was broken if it was -1 or None.

Fix #259

### DIFF
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -131,11 +131,10 @@ class JobSingleton(type):
         # Update alias prefixes
         job._sis_alias_prefixes.add(gs.ALIAS_AND_OUTPUT_SUBDIR)
 
-        # add stacktrace information if not None. if set to -1 use full stack
-        if gs.JOB_ADD_STACKTRACE_WITH_DEPTH is not None and gs.JOB_ADD_STACKTRACE_WITH_DEPTH != 0:
-            stacktrace = traceback.extract_stack(
-                limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0 else None
-            )
+        # add stacktrace information.
+        stack_limit = max(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, 0) if gs.JOB_ADD_STACKTRACE_WITH_DEPTH else 0
+        if stack_limit > 0:
+            stacktrace = traceback.extract_stack(limit=stack_limit)
             job._sis_stacktrace.append(stacktrace)
 
         return job
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -132,9 +132,13 @@ class JobSingleton(type):
         job._sis_alias_prefixes.add(gs.ALIAS_AND_OUTPUT_SUBDIR)
 
         # add stacktrace information.
-        stack_limit = max(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, 0) if gs.JOB_ADD_STACKTRACE_WITH_DEPTH else 0
-        if stack_limit > 0:
-            stacktrace = traceback.extract_stack(limit=stack_limit)
+        assert (
+            isinstance(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, int) or gs.JOB_ADD_STACKTRACE_WITH_DEPTH == float("inf")
+        ) and gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0
+        if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0:
+            stacktrace = traceback.extract_stack(
+                limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if isinstance(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, int) else None
+            )
             job._sis_stacktrace.append(stacktrace)
 
         return job
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -133,7 +133,7 @@ class JobSingleton(type):
 
         # add stacktrace information, if set to None or -1 use full stack
         stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
-        if stack_depth > 1 or None:
+        if stack_depth is None or isinstance(stack_depth, int) and stack_depth > 1:
             # add +1 for the traceback command itself, and remove it later
             stacktrace = traceback.extract_stack(limit=stack_depth)[:-1]
             job._sis_stacktrace.append(stacktrace)
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -42,7 +42,7 @@ def get_args(f, args, kwargs):
     parsed_args = inspect.getcallargs(f, self, *args, **kwargs)
     self = inspect.getfullargspec(f).args[0]
     if self != "self":
-        logging.warning("Deleted obj attribute not named self attribute " "name %s object" % self)
+        logging.warning("Deleted obj attribute not named self attribute name %s object" % self)
     del parsed_args[self]
     return parsed_args
 
@@ -1079,7 +1079,7 @@ class Job(metaclass=JobSingleton):
         global SET_DEFAULT_WARNING_COUNT
         if SET_DEFAULT_WARNING_COUNT < 10:
             logging.warning(
-                "set_default is deprecated, please set the variable manually " "(%s, %s, %s)" % (self, name, value)
+                "set_default is deprecated, please set the variable manually (%s, %s, %s)" % (self, name, value)
             )
             SET_DEFAULT_WARNING_COUNT += 1
         elif SET_DEFAULT_WARNING_COUNT == 10:
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -133,7 +133,9 @@ class JobSingleton(type):
 
         # add stacktrace information if not None. if set to -1 use full stack
         if gs.JOB_ADD_STACKTRACE_WITH_DEPTH is not None and gs.JOB_ADD_STACKTRACE_WITH_DEPTH != 0:
-            stacktrace = traceback.extract_stack(limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0 else None)
+            stacktrace = traceback.extract_stack(
+                limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0 else None
+            )
             job._sis_stacktrace.append(stacktrace)
 
         return job
--- a/sisyphus/job.py
+++ b/sisyphus/job.py
@@ -133,9 +133,8 @@ class JobSingleton(type):
 
         # add stacktrace information, if set to None or -1 use full stack
         stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
-        if stack_depth is None or stack_depth > 1:
-            # add +1 for the traceback command itself, and remove it later
-            stacktrace = traceback.extract_stack(limit=stack_depth)[:-1]
+        if gs.JOB_ADD_STACKTRACE_WITH_DEPTH is not None and gs.JOB_ADD_STACKTRACE_WITH_DEPTH != 0:
+            stacktrace = traceback.extract_stack(limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0 else None)
             job._sis_stacktrace.append(stacktrace)
 
         return job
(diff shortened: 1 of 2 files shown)


sisyphus/job.py
```diff
@@ -133,7 +133,7 @@ def __call__(cls: Type[T], *args, **kwargs) -> T:
 
         # add stacktrace information, if set to None or -1 use full stack
         stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
-        if stack_depth > 1 or None:
+        if stack_depth is None or isinstance(stack_depth, int) and stack_depth > 1:
```
michelwi, line 136: Why do we need the isinstance check? It seems to me that it just unnecessary narrows down the functionality
```suggestion
        if stack_depth is None or stack_depth > 1:
```

Icemole: Yeah, what you propose makes sense, it was just to "assert" that it was an integer.

albertz:
Yes I also wondered about this. Specifically also, what are valid values for `JOB_ADD_STACKTRACE_WITH_DEPTH`. The definition says:
```python
# Add stacktrace information with specified depth, 0 for deactivation, None or -1 for full stack
JOB_ADD_STACKTRACE_WITH_DEPTH = 0
```

And then in the code:
```python
stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
if stack_depth > 1 or None:
    ...
```

This is wrong in multiple ways. I saw this before, but I was actually not sure about the intended valid values. `None` is never possible.

You change this check to:
```python
if stack_depth is None or isinstance(stack_depth, int) and stack_depth > 1:
    ...
```

But that's not enough. `None` is still not possible because of the check `JOB_ADD_STACKTRACE_WITH_DEPTH >= 0`. `None >= 0` will result in `TypeError: '>=' not supported between instances of 'NoneType' and 'int'. But I was not sure whether the documentation is maybe just wrong, and `None` was never intended to be used? (@JackTemaki, I think you wrote this?)

So we either should fix the documentation, or we should allow `None`.

I also don't quite understand the reasoning of `stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1`. Why add `+1` to the stack depth? In the comment below, it says:
```python
# add +1 for the traceback command itself, and remove it later
```
But why? Why to add +1 and then remove it later?

Icemole:
Thanks for the input! I've opened a separate issue (#259) to investigate these issues. For now I suggest to fix the issue in this PR, which would support `-1` as a value. Maybe we can take a look at supporting `None` (if needed) later.

albertz:
> Thanks for the input! I've opened a separate issue (#259) to investigate these issues. For now I suggest to fix the issue in this PR, which would support `-1` as a value. Maybe we can take a look at supporting `None` (if needed) later.

I don't really understand. It should be trivial to fix all of that. It's one line of code to change only?

sisyphus/job.py
```diff
@@ -133,7 +133,7 @@ def __call__(cls: Type[T], *args, **kwargs) -> T:
 
         # add stacktrace information, if set to None or -1 use full stack
         stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
-        if stack_depth > 1 or None:
+        if stack_depth is None or stack_depth > 1:
             # add +1 for the traceback command itself, and remove it later
             stacktrace = traceback.extract_stack(limit=stack_depth)[:-1]
```
albertz, line 138: So I personally would go with changing the meaning of `JOB_ADD_STACKTRACE_WITH_DEPTH = None` to handle this as disabled.
```suggestion
        if gs.JOB_ADD_STACKTRACE_WITH_DEPTH is not None and gs.JOB_ADD_STACKTRACE_WITH_DEPTH != 0:
            stacktrace = traceback.extract_stack(limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0 else None)
```

sisyphus/job.py
```diff
@@ -133,7 +133,7 @@ def __call__(cls: Type[T], *args, **kwargs) -> T:
 
         # add stacktrace information, if set to None or -1 use full stack
         stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
```
albertz, line 135:
```suggestion
        # add stacktrace information if not None. if set to -1 use full stack
```

albertz: And also change the doc in the global settings.

albertz:
> Yes, I wanted to avoid too much branching, but sure, we can do it here as well.

I don't see that there is more branching/code than before. With my proposed changes, there is even less code/logic than before, or not? It removes the whole extra variable, and the logic with depth+1 and then depth-1 again.

sisyphus/job.py
michelwi: if so, then lets remember to also change the default value in https://github.com/rwth-i6/sisyphus/blob/e61ba9ffb54e0fb421cc28d7c37457971663ed29/sisyphus/global_settings.py#L355

albertz: 0 also still means disabled with the proposed code change. So both 0 and None means deactivated.

Icemole: I checked internally the workings of `traceback.extract_stack`. Your implementation seems correct :+1:

Icemole:
> branching/code

I meant "discussions" rather than "if statements" :slightly_smiling_face:

sisyphus/job.py
albertz: No `extract_stack(limit=None)` means to get the full stack. I think `extract_stack(limit=-1)` is not valid.

albertz:
You still need to update the doc about the setting. None means disabled now.

albertz:
(I personally would keep this PR on topic only about the stack depth, not about random other formatting fixes. The other Ruff PR would fix those anyway.)

sisyphus/job.py
michelwi: > 0 also still means disabled with the proposed code change. So both 0 and None means deactivated.

Right.. but should we then just drop None from the permissible values and simply it even more?

> extract_stack(limit=-1) is not valid.

Negative numbers will print from the back of the stack
https://docs.python.org/3.10/library/traceback.html#traceback.extract_stack

Which brings me to another approach: just pass on the value of gs.JOB_ADD_STACKTRACE_WITH_DEPTH if it is different from 0 and then any negative number is allowed.
This would redefine the meaning of -1, keep the meaning 0 and None, and allow for more flexibility.

albertz: It just feels a bit weird/confusing to me that the meaning of 0 is somehow the opposite of the meaning of None. But I don't really have strong opinions here.

The question is also, what have people used so far? We should try to avoid breaking existing setups. But I guess noone has used None, as that was broken anyway. The same for -1. So I just have set this to a high number (100) to enable it with full stack. But none of the discussed options would break any of that, so I guess we can decide for whatever we want.

albertz: @Icemole You still did not consider this here. But there was also no clear conclusion here yet. Maybe we can really just not allow `None` for now, and require `int`.

albertz:
> Please let me know if the changes don't suit you. Thanks!

You ignored the [open thread above](https://github.com/rwth-i6/sisyphus/pull/258#discussion_r2078233621).

There was still no clear conclusion here.

Currently I tend to think that the following makes sense:

- None: Disallow. None was never possible so far, and the correct meaning is somehow ambiguous. None makes sense for me to mean disabled (just like 0). But it could also mean to have no limit for the stack. Having it disallowed now also allows to redefine it later.
- 0: Keep current logic: Disable the extraction.
- `>=1`: Keep current logic. Extraction with limit.

But then, I'm less sure about other values. For -1 (or other negative values), we could interpret this as extracting without limit. Or we can overtake the `extract_stack` logic for it, as @michelwi proposed. I'm just not sure how useful this is. Has anyone ever wanted to have it like this? However we define this now, it's difficult to change later, so we should think about it. Or we just disallow this as well for now, as this is also ambiguous.

We can also allow `float("inf")` or so, to mean no limit. That would be quite explicit and non-ambiguous.

In any case, it can also be extended later by other values, maybe strings, or dict with options, to define some more custom logic.

So, to conclude, my suggestion:

- `<0`: Disallow for now, as this is also ambiguous.
- `float("inf")`: Extract stack without limit.

sisyphus/job.py
Icemole: Thanks for the heads-up.

The issue is that, sadly, `depth=None` doesn't have the same semantical meaning as `limit=None`, so I wouldn't use `None`. It anyway has the same meaning as `0`.

I would personally only allow positive numbers (and zero). I also find `None` and `-1` a bit confusing, especially given the difference between "depth" and "limit".

How about the following?

```
stack_limit = max(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, 0) if gs.JOB_ADD_STACKTRACE_WITH_DEPTH else 0
stacktrace = traceback.extract_stack(limit=stack_limit)
```

Icemole:
Yes, I agree to this implementation. I find `float("inf")` clearer than `-1` to retrieve the full stack.

sisyphus/job.py
albertz: Before we discuss the implementation (which is really trivial), let us discuss and agree on some wanted behavior. I summarized some potential behavior for all the possible cases [below](https://github.com/rwth-i6/sisyphus/pull/258#issuecomment-2871609864). Does everyone agree to follow that?

albertz:
Ok, does @michelwi agree as well to this? If so, then we can discuss the code change (I have a couple of comments on that).

sisyphus/job.py
albertz: > you suggested `-1` to be used as full stack as well.

No, this is outdated. The post [below](https://github.com/rwth-i6/sisyphus/pull/258#issuecomment-2871609864) is newer than that. Let's continue the discussion there (I moved it there because you seemed to have missed it here...). I think I discuss all possible cases there. Or do I miss anything?

Icemole:
I just noticed the difference between "disallow" and "set to 0". I slightly prefer setting `None` and a value `< 0` to `0` without annoying the user, but that's just me. Anyway I'm also ok with disallowing these values, as it's only a one-time change and we've already changed the documentation (and also `-1` and `None` were broken before, so nobody could have been using that).

albertz:
> I slightly prefer setting `None` and a value `< 0` to `0` without annoying the user, but that's just me

Why? We just agreed and discussed before that both None and `<0` are ambiguous in the meaning (you could maybe interpret `None` as having a stack without limit; or negative numbers are starting the stack from the back). What value does it give you to have other ways to disable the stack extraction, when you already have 0 for that?

michelwi:
> Anyway, having them disallowed for now gives us the possibility to also change the meaning of them at some later point.

strongly agree to this.

I just have a small negative feeling of
> and require int. [for JOB_ADD_STACKTRACE_WITH_DEPTH]

vs.
> float("inf"): Extract stack without limit.

not being an integer. But there is no integer inf and out of all the special values to represent "no limit" float inf is as good to me as any other. Or has anyone a much nicer solution?

albertz:
Ok, I think we agree then to the last suggestion here, and there are no other suggestions regarding `float("inf")`.

So then we can now discuss on the specific implementation for that. I will review the code.

sisyphus/job.py
```diff
@@ -131,11 +131,10 @@ def __call__(cls: Type[T], *args, **kwargs) -> T:
...
 
-        # add stacktrace information, if set to None or -1 use full stack
-        stack_depth = gs.JOB_ADD_STACKTRACE_WITH_DEPTH + 1 if gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 else None
-        if stack_depth > 1 or None:
-            # add +1 for the traceback command itself, and remove it later
-            stacktrace = traceback.extract_stack(limit=stack_depth)[:-1]
+        # add stacktrace information.
+        stack_limit = max(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, 0) if gs.JOB_ADD_STACKTRACE_WITH_DEPTH else 0
+        if stack_limit > 0:
+            stacktrace = traceback.extract_stack(limit=stack_limit)
```
albertz, line 137:
```suggestion
        # add stacktrace information.
        assert gs.JOB_ADD_STACKTRACE_WITH_DEPTH >= 0 and (isinstance(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, int) or gs.JOB_ADD_STACKTRACE_WITH_DEPTH == float("inf"))
        if gs.JOB_ADD_STACKTRACE_WITH_DEPTH > 0:
            stacktrace = traceback.extract_stack(limit=gs.JOB_ADD_STACKTRACE_WITH_DEPTH if isinstance(gs.JOB_ADD_STACKTRACE_WITH_DEPTH, int) else None)
```

Icemole: We should change the order of the conditions so that `isinstance` is checked before doing `>=`, so that we don't have any types incompatible with `>=`. Other than that, looks good!

Icemole: Fixed in 71d078e54bd00be1b6de7d145c5883081c6748f9.

Resolving thread.

albertz (approved):

michelwi (approved):

PR #261, opened 2025-05-08
Change formatter to ruff

michelwi:
As discussed offline today the ruff formatter is faster, as easy to use, and yields equal or better results compared to black. 
We therefore want to update the formatter to ruff.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,4 +1,4 @@
-name: black
+name: ruff
 
 on:
   push:
@@ -11,7 +11,7 @@ on:
       - master
 
 jobs:
-  check-black-formatting:
+  check-ruff-formatting:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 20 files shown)


.github/workflows/ruff.yml
```diff
@@ -11,15 +11,14 @@ on:
...
     - uses: actions/setup-python@v5
       with:
         python-version: 3.13
         cache: 'pip'
-        cache-dependency-path: '.github/workflows/black.yml'
-    - run: pip install black==22.3.0
-    - run: black --diff .
-    - run: black --check .
+        cache-dependency-path: '.github/workflows/ruff.yml'
+    - run: pip install ruff==0.9.7
```
albertz, line 23: It seems we are not really consistent here across our repos. E.g. @NeoLegends in RETURNN used a different Ruff version. Why do we do this? I think it would be better if we use the same Ruff version everywhere.

michelwi: > I think it would be better if we use the same Ruff version everywhere.

agreed.

> Why do we do this?

I copy pasted the version we (already) use in the apptek_asr repo to be consistent. I guess the exact version is determined by "latest version at the time of switching to ruff".

curufinwe: Let's switch to the current version (0.11.8). We can do the switch in apptek_asr later once I build a new runtime.

michelwi: I just tried locally and there would be no file reformatted when changing between versions 0.9.7 and 0.11.8 for sisyphus (and apptek_asr as well).

> Let's switch to the current version (0.11.8). 

will do so now.

curufinwe (approved):

albertz (approved):

NeoLegends (approved):
🚀

PR #264, opened 2025-06-03
Handle bound methods in tools.extract_paths

dthulke:
With this change paths from the instance a bound method is bound to are also extracted when calling `tools.extract_paths`.

Specifically, I needed this as my custom worker_wrapper depends on a job that builds the venv that is used in the worker_wrapper.

I'm unsure whether it's better to implement this here or in `get_object_state`. In `get_object_state` this might influence the hashes of existing setups (but if I see it correctly bound methods are not handled properly there at the moment).

### DIFF
--- a/sisyphus/tools.py
+++ b/sisyphus/tools.py
@@ -15,6 +15,7 @@ import logging
 import subprocess
 import linecache
 from typing import Set, Any
+from inspect import ismethod
 
 try:
     import tracemalloc
@@ -99,6 +100,8 @@ def extract_paths(args: Any) -> Set:
             for k, v in obj.items():
                 if not isinstance(k, str) or not k.startswith("_sis_"):
                     queue.append(v)
+        elif ismethod(obj) and hasattr(obj, "__self__") and obj.__self__ is not None:
+            queue.append(obj.__self__)
         else:
             queue.append(get_object_state(obj))
     return out


dthulke:
Just to clarify how the worker wrapper would look like:

```python
class VenvRuntime:
  def __init__(self, venv_path: Path):
    self.venv_path = venv_path
    
  def worker_wrapper(self, job: tk.Job, task_name: str, call: List[str]):
    # Ingest the venv_path in call 
```
and then in the manager:
```python
venv_path = CreateVenvJob(...)
venv_runtime = VenvRuntime(venv_path.out)

gs.worker_wrapper = venv_runtime.worker_wrapper
my_job = MyJobRunningInVenvJob(...)
```

I could change this, but wanted to keep consistency with existing setups in how the worker_wrapper is defined.

> But then Sisyphus completely ignores the worker_wrapper when determining the dependencies of a Job, so I dont quite see yet how having the worker_wrapper as a bound method helps here.

Yes, this requires an addition change in the job initialisation by adding the worker wrapper here (should not have any negative side effect I guess):
https://github.com/rwth-i6/sisyphus/blob/d1c34cf0593884705715b814a030edb94c0f6bb7/sisyphus/job.py#L237

(first I thought that it's included via dict but tools.extract_paths removes everything starting with _sis ...)

> [Alternatives suggested]

The alternatives would also work would also work but would all require more changes to existing setups. The suggested implementation would work directly with existing setups that use the previous workflow to set the worker wrapper.

dthulke:
An orthogonal question is I guess what the expected behaviour in the example you shared is:

> ```
> class MethodProvider:
>     def __init__(self, some_path: tk.Path):
>         self.path = some_path
> 
>     def some_method(self):
>         [do something based on self.path]
> 
> class SomeJob(Job):
>     def __init__(self, method):
>         self.method = method
> 
>     def run(self):
>         self.method()
> 
> def py():
>     my_path = CreateSomethingJob().out_path
>     my_method_provider = MethodProvider(my_path)
>     my_job = SomeJob(my_method_provider.some_method)
> ```

1. Do we want to allow using (bound) methods as job inputs?
2. If not, shouldn't we throw an error?
3. If yes, then in your example SomeJob should also depend on the output of CreateSomethingJob otherwise this would break the dependencies (this PR)
4. Another question is then further whether the hash of SomeJob should also depend on CreateSomethingJob (https://github.com/rwth-i6/sisyphus/pull/265)

michelwi:
> Just to clarify how the worker wrapper would look like:
[...]

Thanks for the clarification! I agree that we have two orthogonal problems here
1. Should bound methods be allowed as Job inputs and how would we implement it.
2. should the worker_wrapper be considered for dependencies of the Jobs.

I think that for 1. most problems can be avoided if the class instance is passed instead of its methods. Otherwise there can be all kinds of ambiguities.. (e.g. one class has a staticmethod that shall be used and you pass it and not the next Job has a lot of unnecessary dependencies of the specific instance you happened to grab the method from).. so I would hope to avoid this discussion... But anyway, this is not your problem here.

For 2. I am a bit hesitant, as the worker_wrapper is a "global setting" and it is already is a stretch beyond the original idea that we change it mid setup. Now if we also include it as a part of the computational graph this leads to me asking the question "where does it end?" should we now also consider the gs.engine for dependencies ("But I want to run my InstallSlurmJob and then update the engine to use the newly installed slurm version") or any environment variable passed to the Job?
My world view would be much saner if the settings remained separate from the graph construction.

> > [Alternatives suggested]
>
> The alternatives would also work but would all require more changes to existing setups. The suggested implementation would work directly with existing setups that use the previous workflow to set the worker wrapper.

I think especially the implementation with the context manager (you may hide this class even inside some other module) would be as non-intrusive as it gets and only requires one additional line in your main config, which you anyway already need to modify to add the CreateVenvJob. e.g.
```python
from existing_setups.legacy import legacy_workflow
from helpers import WorkerWrapperHandler

def py():
    venv_path = CreateVenvJob(...).out_venv
    with WorkerWrapperHandler(venv_path):
        legacy_workflow()
```
should correctly add the dependencies on the CreateVenvJob to any Job instanciated inside the legacy workflow without the need to change its implementation.

dthulke:
> one class has a staticmethod that shall be used and you ...

That's not an issue - a static method is not a method but a function (according to inspect.ismethod / isfunction).

But I agree, let's ignore 1.

>  I am a bit hesitant, as the worker_wrapper is a "global setting" and it is already is a stretch beyond the original idea that we change it mid setup.

By allowing to change the worker wrapper mid setup it is not really a global setting anymore - so as the worker wrapper is anyway in many setups already defined inside the config in my opinion it makes sense to have it also part of the computation graph.

> "where does it end?"

My first thought would be: anything defined and changed in the config should be part of the computation graph.

If we allow to change the engine for some jobs, why not add a job that builds the dependencies for that engine?

One additional question is then whether the worker_wrapper should influence the hash of jobs. One argument for this is that if you e.g. create two venvs with the same packages in two completely different singularity images, you actually want an image specific python venv. If the path to the image is not part of the hash, you will only get a single python venv for both images what might cause issues...

But yes, one of the suggested alternatives should work fine.

@curufinwe any opinions (we briefly talked about this)?

dthulke:
Closing this as we decided not to include the worker wrapper in the graph per default.

PR #273, opened 2025-09-24
DelayedOps: dict.__getitem__ with delayed key

patrick-wilken:
To do something like:
```
models = {1: model_eopch1, 2: model_epoch2, ...}

best_epoch = SelectBestEpochJob(all_scores).out  # tk.Variable

best_model = tk.Delayed(models)[best_epoch]
```

Or just any situation where a dict key depends on a job output. Actually, should also work for list indexing.
`DelayedSlice` already supports both delayed iterable as well as delayed range start/end/step, so this is just the analogous thing for "simple" access.

### DIFF
--- a/sisyphus/delayed_ops.py
+++ b/sisyphus/delayed_ops.py
@@ -139,7 +139,7 @@ class DelayedRound(DelayedBase):
 
 class DelayedGetItem(DelayedBase):
     def get(self):
-        return try_get(self.a)[self.b]
+        return try_get(self.a)[try_get(self.b)]
 
 
 class Delayed(DelayedBase):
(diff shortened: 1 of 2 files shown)


patrick-wilken:
While thinking about what could have made the check fail, I noticed that _theoretically_ this breaks code that already uses delayed dicts (i.e. `tk.Variables` with a dict value, or something like `Delayed(my_dict)`) which at the same time have delayed objects as keys. Because now these keys would be resolved to the delayed value.
Do you think someone is that evil?
I could imagine that somewhere there exist dicts with tk.Paths as keys, but probably not in combination with the above.
I think the whole delayed thing can only be unambiguous if we expect none of the operations that eventually run to use delayed objects as operands. For example, `delayed_var1.fallback(delayed_var2)` could as fallback return `delayed_var2` itself or the underlying value, we choose the latter. Given that assumption, the changes here should be okay.

PR #302, opened 2026-06-11
Print job complete percentage in manager (if available)

dorian-K:
image

### DIFF
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -299,10 +299,9 @@ class Manager(threading.Thread):
                 job_manager_info_string = job.info()
                 if job_manager_info_string is not None:
                     info_string += " {%s}" % job_manager_info_string
-            if hasattr(job, "completed_fraction"):
-                frac = job.completed_fraction()
-                if frac is not None:
-                    info_string += " [%.1f%%]" % (frac * 100)
+            frac = job.completed_fraction()
+            if frac is not None:
+                info_string += " [%.1f%%]" % (frac * 100)
 
         if state == gs.STATE_RUNNING and gs.PRINT_OLD_LOG_FILE_RUNNING is not None:
             task = job._sis_next_task()
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -7,16 +7,20 @@ import asyncio
 import atexit
 import logging
 import os
+import re
 import sys
 import threading
 import time
 import traceback
 import warnings
+from ast import literal_eval
+from datetime import datetime
 from typing import TYPE_CHECKING, Dict, Collection
 
 from multiprocessing.pool import ThreadPool
 
 from sisyphus import toolkit, tools
+from sisyphus.logging_format import color_mapping as _log_level_color
 from sisyphus.loader import config_manager
 from sisyphus.block import Block
 from sisyphus.tools import finished_results_cache
@@ -24,6 +28,8 @@ import sisyphus.global_settings as gs
 
 if TYPE_CHECKING:
     from sisyphus.job import Job
+
+_LOG_TIMESTAMP_RE = re.compile(r"\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2})")
 
 
 class JobCleaner(threading.Thread):
@@ -302,7 +308,48 @@ class Manager(threading.Thread):
             if hasattr(job, "completed_fraction"):
                 frac = job.completed_fraction()
                 if frac is not None:
-                    info_string += " [%.1f%%]" % (frac * 100)
+                    eta = None
+                    if 0.01 < frac < 0.99:
+                        task = job._sis_next_task()
+                        if task:
+                            rqmt_secs = None
+                            try:
+                                line = None
+                                with open(task.path(gs.ENGINE_SUBMIT)) as f:
+                                    for line in f:
+                                        pass
+                                if line:
+                                    _, rqmt = literal_eval(line)
+                                    t = rqmt.get("time")
+                                    if t:
+                                        rqmt_secs = float(t) * 3600
+                            except Exception:
+                                pass
+                            log_file = job._sis_path(gs.JOB_LOG + "." + task.name(), task.task_ids()[0])
+                            if os.path.isfile(log_file):
+                                try:
+                                    with open(log_file, "rb") as f:
+                                        first_line = f.readline().decode("ascii", errors="ignore")
+                                    m = _LOG_TIMESTAMP_RE.match(first_line)
+                                    if m:
+                                        start_ts = datetime.strptime(m.group(1), "%Y-%m-%d %H:%M:%S").timestamp()
+                                        elapsed = time.time() - start_ts
+                                        if elapsed > 120:
+                                            remaining = int(elapsed * (1 - frac) / frac)
+                                            if remaining > 99 * 3600:
+                                                eta = ">99h"
+                                            else:
+                                                h, rem = divmod(remaining, 3600)
+                                                mn = rem // 60
+                                                eta = f"~{h}h{mn:02d}m" if h else f"~{mn}m"
+                                            if rqmt_secs and remaining > rqmt_secs:
+                                                eta = f"\x1b[31m{eta}{_log_level_color(logging.INFO)}"
+                                except Exception:
+                                    pass
+                    if eta is not None:
+                        info_string += f" [{frac * 100:.1f}% done, ETA {eta}]"
+                    else:
+                        info_string += " [%.1f%%]" % (frac * 100)
 
         if state == gs.STATE_RUNNING and gs.PRINT_OLD_LOG_FILE_RUNNING is not None:
             task = job._sis_next_task()
--- a/sisyphus/manager.py
+++ b/sisyphus/manager.py
@@ -294,10 +294,15 @@ class Manager(threading.Thread):
 
     @staticmethod
     def print_job_state(state, job, info_string, verbose=False):
-        if hasattr(job, "info") and state == gs.STATE_RUNNING:
-            job_manager_info_string = job.info()
-            if job_manager_info_string is not None:
-                info_string += " {%s}" % job_manager_info_string
+        if state == gs.STATE_RUNNING:
+            if hasattr(job, "info"):
+                job_manager_info_string = job.info()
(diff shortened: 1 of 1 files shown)


sisyphus/manager.py
```diff
@@ -294,10 +294,15 @@ def get_job_info_string(state, job, verbose=False):
...
-        if hasattr(job, "info") and state == gs.STATE_RUNNING:
-            job_manager_info_string = job.info()
-            if job_manager_info_string is not None:
-                info_string += " {%s}" % job_manager_info_string
+        if state == gs.STATE_RUNNING:
+            if hasattr(job, "info"):
+                job_manager_info_string = job.info()
+                if job_manager_info_string is not None:
+                    info_string += " {%s}" % job_manager_info_string
+            if hasattr(job, "completed_fraction"):
```
albertz: I wonder a bit about the `hasattr`. Jobs should always have that method. But we also have the same weird check for `info` above. I guess it doesn't hurt.

dorian-K: Maybe if an old job version was pickled to disk and then read, maybe it wouldn't have that function? But I'm not sure
claude code agrees that they are unnecessary

dorian-K:
image
Added ETA until job finishes running, with red text when the eta surpasses the job's time rqmt

sisyphus/manager.py
albertz: I think this function here is only called in the SIs manager?

sisyphus/manager.py
```diff
@@ -294,10 +300,56 @@ def get_job_info_string(state, job, verbose=False):
...
+                                line = None
+                                with open(task.path(gs.ENGINE_SUBMIT)) as f:
+                                    for line in f:
+                                        pass
+                                if line:
+                                    _, rqmt = literal_eval(line)
+                                    t = rqmt.get("time")
+                                    if t:
+                                        rqmt_secs = float(t) * 3600
+                            except Exception:
```
albertz: I don't like such generic `except Exception: pass`.

sisyphus/manager.py
```diff
@@ -294,10 +300,56 @@ def get_job_info_string(state, job, verbose=False):
...
+                                            remaining = int(elapsed * (1 - frac) / frac)
+                                            if remaining > 99 * 3600:
+                                                eta = ">99h"
+                                            else:
+                                                h, rem = divmod(remaining, 3600)
+                                                mn = rem // 60
+                                                eta = f"~{h}h{mn:02d}m" if h else f"~{mn}m"
+                                            if rqmt_secs and remaining > rqmt_secs:
+                                                eta = f"\x1b[31m{eta}{_log_level_color(logging.INFO)}"
+                                except Exception:
```
albertz: Same here.

sisyphus/manager.py
```diff
@@ -294,10 +300,56 @@ def get_job_info_string(state, job, verbose=False):
...
+                                        if elapsed > 120:
+                                            remaining = int(elapsed * (1 - frac) / frac)
+                                            if remaining > 99 * 3600:
+                                                eta = ">99h"
+                                            else:
+                                                h, rem = divmod(remaining, 3600)
+                                                mn = rem // 60
+                                                eta = f"~{h}h{mn:02d}m" if h else f"~{mn}m"
+                                            if rqmt_secs and remaining > rqmt_secs:
+                                                eta = f"\x1b[31m{eta}{_log_level_color(logging.INFO)}"
```
albertz: I'm not sure on this. At RZ or FZJ, we have this case quite frequently that our training jobs are only submitted with a 12h or 24h limit but trainings run much longer. Now with the `completed_fraction` and better Sisyphus resume logic, and also with `stop_for_resubmission_when_low_time_left` on RETURNN side, it's not really an issue anymore. But the red color suggests that there is an issue.

michelwi:
I want to raise the question "should we really do this?"
* I only took a coursory glance, but it seems to me that for each running job we need to open and parse 2 or 3 files. I already foresee problems for people that have 1k+ running jobs on a slow filesystem day. 
* Also this update is printed quite often (when I have so many running jobs, i tend to get an update every other minute)
* Not sure how resilient your implemented logic is, but there seem to be a lot of safeguards necessary (`if a: [program continues]` or `try: ... Except`) Are we sure this does not break easily?
* The implementation has huge levels of indentation and is not easy to read. I would suggest to transform many parts of the form
 ```
if a:
    [expected case, long code follows]
```
into
```
if not a:
    [Handle problem; e.g. pass, continue, or set a default]
[continue normal execution]
```
which improves readability. Maybe even put some code in a separate function and then just `return ""` in case of problems.

And all of this just for the tiny additional information about some estimated finish time (which I think is irrelevant as I anyway cannot change it and the job finishes when it does :shrug: )

**Comment by albertz** [comment]
> I only took a coursory glance, but it seems to me that for each running job we need to open and parse 2 or 3 files.

No. Not more files than before are being opened. This is only for jobs which support the `completed_fraction`, which is mostly just the `ReturnnTrainingJob` currently. And already right now it opens those files to print the progress in what epoch it currently is. This here basically just extends that, and works in general on any other job that supports the `completed_fraction`.

You are right that the LR file that is opened by `ReturnnTrainingJob` is now opened twice right after each other due to first calling `info` and then `completed_fraction`. We could maybe cache this, but the OS level cache might also handle that well already.

So I'm not sure this is really such an issue.

File: sisyphus/manager.py [hunk 5]
```diff
@@ -294,10 +300,56 @@ def get_job_info_string(state, job, verbose=False):
...
+                            if os.path.isfile(log_file):
+                                try:
+                                    with open(log_file, "rb") as f:
+                                        first_line = f.readline().decode("ascii", errors="ignore")
+                                    m = _LOG_TIMESTAMP_RE.match(first_line)
+                                    if m:
+                                        start_ts = datetime.strptime(m.group(1), "%Y-%m-%d %H:%M:%S").timestamp()
+                                        elapsed = time.time() - start_ts
+                                        if elapsed > 120:
+                                            remaining = int(elapsed * (1 - frac) / frac)
```
michelwi: This assumes that all of the progress of the job (`frac`) was made in the current engine submission (slurm task), since only the current logfile is read to calculate elapsed.
If a job is resumable and can carry over progress between submissions (e.g. the ReturnnTrainingJob, which is the only job that is currently supported by this code) and had been interrupted at least once, then this estimate is just wrong. 

We would have to read all of the logfiles (or the usage file); but even this could be wrong, if done progress is lost between submissions (e.g. the last incomplete sub-epoch of a ReturnnTrainingJob if stop_for_resubmission_when_low_time_left is not used)

sisyphus/manager.py
```diff
@@ -294,10 +300,56 @@ def get_job_info_string(state, job, verbose=False):
...
+                                    t = rqmt.get("time")
+                                    if t:
+                                        rqmt_secs = float(t) * 3600
+                            except Exception:
+                                pass
+                            log_file = job._sis_path(gs.JOB_LOG + "." + task.name(), task.task_ids()[0])
+                            if os.path.isfile(log_file):
+                                try:
+                                    with open(log_file, "rb") as f:
+                                        first_line = f.readline().decode("ascii", errors="ignore")
```
michelwi: I think only reading the first line of the ligfile and probably also the regex used to get the timestamp is specific to the slurm engine and then also to your specific slurm installation that sets the Prolog in this way. 

This code should work for all supported engines.

michelwi (changes requested):

dorian-K:
I've read through your comments and i agree with most of them. I actually already noticed a slight slowdown with only 30 running jobs, as before the print was instant, and now it printed line-by-line for maybe 100-200ms until it was done. So for thousands of jobs this may get super annoying.

This and the slurm-specific issue could likely be improved by using the file's metadata ("birthtime"), but this is, at the moment, not easily accessible in python without external libraries or subprocess calls. And anyways it would still be a crude estimate and be wildly inaccurate for interrupted/rescheduled jobs as you've said above.

So i have decided to remove the ETA thing again and only keep the percentage. Probably better to implement this feature individually for some (long running) jobs in their `info()` to get more accurate numbers

albertz:
> file's metadata ("birthtime")

Python 3.12+ populates this on compatible Linux systems `os.stat(path).st_birthtime`?

Or Python 3.11 and Older:
```python
info = os.statx(os.AF_DECNET, path, mask=os.STATX_BIRTHTIME)
    
# Check if the filesystem actually returned the birthtime
if info.stx_mask & os.STATX_BIRTHTIME:
    birthtime = info.stx_btime.tv_sec
```

dorian-K:
```
❯ ./.venv/bin/python
Python 3.12.13 (main, Mar 20 2026, 00:33:26) [Clang 22.1.1 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> a = os.stat("work/i6_experiments/users/dorian_koch/speech_llm/moshi/MoshiAnnotate.Y6urTHFSgFbE.cleared.0001/log.run.1")
>>> a.st_birthtime
  File "<stdin>", line 1, in <module>
AttributeError: 'os.stat_result' object has no attribute 'st_birthtime'
>>> a
os.stat_result(st_mode=33184, st_ino=324265055682040913, st_dev=1990885514, st_nlink=2, st_uid=46724, st_gid=50526, st_size=2916, st_atime=1781262983, st_mtime=1781240318, st_ctime=1781240318)
>>> os.statx
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'statx'. Did you mean: 'stat'?
```

dorian-K:
The command `stat` does give the birthtime though, so not an os issue

albertz:
Ok I guess then we would need a better way to get the job start time. Maybe extending the engine API or so. Or storing it explicitly in a file when the job starts.

dorian-K:
> Ok I guess then we would need a better way to get the job start time. Maybe extending the engine API or so. Or storing it explicitly in a file when the job starts.

Yes. But in the case of interrupted jobs that need multiple submissions to finish (i.e. training jobs), this would again be inaccurate, just like the previous approach. Perhaps we should leave it for now

But anyways, I believe this pull request is ready to be merged?

michelwi (approved):
> I believe this pull request is ready to be merged?

yes, we can merge as is and work on a better implementation for the ETA later.

PR #304, opened 2026-06-24
OutputReport, atomic write

Zettelkasten:
In out setup, writing reports can take some time (especially directly after startup, when the file system is not cached yet and the reports have a lot of inputs).
Previously I had the issue that this meant that the report file on disk would first be re-created (emptied), then the report is built (which takes up to a second when the file system is having a bad day), and only after the final report is written to disk again.

This fixes this to first write to a temp file, and then do an atomic swap once it's ready so that the output report should always exist.

### DIFF
--- a/sisyphus/graph.py
+++ b/sisyphus/graph.py
@@ -16,6 +16,7 @@ import threading
 from typing import DefaultDict, Optional, List
 from multiprocessing.pool import ThreadPool
 from datetime import datetime
+import tempfile
 
 
 class Node(object):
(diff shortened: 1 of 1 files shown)


sisyphus/graph.py
```diff
@@ -196,23 +197,29 @@ def write_report(self):
...
                 os.makedirs(outfile_dir)
 
-            # Remove link to avoid overwriting other files
-            if os.path.islink(outfile_name):
-                os.unlink(outfile_name)
-
-            # Actually write report
-            with open(outfile_name, "w") as f:
+            # Write report to temp file first
+            fd, tmp_name = tempfile.mkstemp(dir=outfile_dir)
```
michelwi, line 206: use `gs.TMP_PREFIX` as prefix for the file

sisyphus/graph.py
```diff
@@ -196,23 +197,29 @@ def write_report(self):
...
+            with os.fdopen(fd, "w") as f:
                 if self._report_template:
                     f.write(self._report_template.format(**self._report_values))
                 elif callable(self._report_values):
                     f.write(str(self._report_values()))
                 else:
                     f.write(pprint.pformat(self._report_values, width=140) + "\n")
+
+            # Atomically replace target after temp file is complete to avoid incomplete reports
+            os.replace(tmp_name, outfile_name)
```
michelwi, line 216: > The operation may fail if src and dst are on different filesystems.

https://docs.python.org/3/library/os.html#os.replace

I think this will be an issue since the tmpfile would reside on node-local storage and the output should live on the fileserver?

michelwi (changes requested):

sisyphus/graph.py
Zettelkasten: Yes, that's why I need to already write the temporary file onto the same file system as the final destination: the `os.replace` needs to be atomic so it cannot copy between different file systems. Otherwise I have the same issue as before that I cannot read the report in the moment that it's copied between file systems.
This also implies that I cannot use `gs.TMP_PREFIX` as you suggested above.

michelwi: mh.. right.. `dir=outfile_dir` guarantees to be on the same filesystem. ok.

sisyphus/graph.py
```diff
@@ -196,23 +197,29 @@ def write_report(self):
...
         except IOError as e:
+            if tmp_name is not None:
+                try:
+                    os.unlink(tmp_name)
+                except OSError:
+                    pass
```
michelwi, line 225: would you consider moving this block into a `finally` block?

michelwi (approved):

albertz (approved):

PR #308, opened 2026-08-10
Add tk.running_in_manager(), set for manager and console

Stefanwuu:
Add a running_in_manager() function to tell whether the code is running in the manager or runtime. Added because the counterpart running_in_worker() would return False in a Returnn training job, where RETURNN runs in a new Python subprocess where Sisyphus is imported again.


michelwi (approved):
lgtm, I assume you tested it and it works better than the old running_in_worker check.

Stefanwuu:
> lgtm, I assume you tested it and it works better than the old running_in_worker check.

Yes, I tested it, and it works.

PR #312, opened 2026-09-09
JOB_INFO_USED_BY_BASE_DIR - store used-by-base-dir in info

albertz:


sisyphus/job.py
```diff
@@ -296,11 +300,19 @@ def _sis_setup_directory(self, force=False):
...
+                if gs.JOB_INFO_USED_BY_BASE_DIR:
+                    _job_info_add_used_by_base_dir(creator)
 
         # export the actual job
         with gzip.open(self._sis_path(gs.JOB_SAVE), "w") as f:
             pickle.dump(self, f)
 
+        used_by_base_dirs = []
+        if gs.JOB_INFO_USED_BY_BASE_DIR:
+            used_by_base_dirs = _job_info_read_used_by_base_dirs(self._sis_path(gs.JOB_INFO))
```
michelwi, line 312: Since we are in `Job._sis_setup_directory`, I would have assumed this code runs when initially setting up the Job folder and therefore that the info file does not exist yet. 

In which cases does `_job_info_read_used_by_base_dirs` find something for you?

sisyphus/job.py
```diff
@@ -296,11 +300,19 @@ def _sis_setup_directory(self, force=False):
                 link_name = os.path.join(self._sis_path(gs.JOB_INPUT), str(job_id).replace("/", "_"))
                 if not os.path.exists(link_name) and not os.path.islink(link_name):
                     os.symlink(src=os.path.abspath(str(creator._sis_path())), dst=link_name, target_is_directory=True)
+                if gs.JOB_INFO_USED_BY_BASE_DIR:
+                    _job_info_add_used_by_base_dir(creator)
```
michelwi, line 304: This will only add the usage indicator to the immediate parents of the Job that is being created. So in a pipeline of e.g.
Data Preparation -> Training -> Recognition

and a different setup now also does some additional recognition, then the TrainingJob will be additionally flagged, but the data preparation jobs will remain unflagged although they are also being reused (== are part of the graph) in a different config.

I think all of the ancestors should be flagged. I was thinking in the direction of "on startup, the manager flags all existing jobs in the current graph and then only the newly created ones", but this would make startup slow every time. 
On the other hand the current approach is suboptimal in the case of data preparation consists of tens of thousands of jobs that create the HDF Files and then we reuse them for multiple training runs -- now each new training job needs to flag all data processing jobs again.

Not sure what a good solution would be.

sisyphus/job.py
albertz: `_sis_setup_directory` is called in many places. When `_sis_setup_since_restart` is False, which is the case at a new start of the Sis mgr, it will redo the setup. This is for example the case for any job it resumes or so.

sisyphus/job.py
albertz: Well, not even all ancestors might exist. E.g. on RZ/FZJ cluster, I only have all the direct ancestors available, but not all the way back to the root. For example, I don't have the `DownloadLibriSpeechCorpusJob` because it should not be on RZ/FZJ filesystem because it has so many small files.

Also, I definitely don't want to iterate through the whole deep ancestors here. It would be too slow. And also way too many files to check. I don't think this is realistic.

michelwi: But are you not concerned that you might end up in a state where the info has some values for USED_BY_BASE_DIR set but some values are missing because it is an indirect dependency? 
That would almost defeat the point of having this info there.

albertz: All setup base dirs are listed that make direct use of the job output. If a setup base dir is not listed, it means it doesn't use it (directly). That is mostly what I care about for cleaning up. It tells me what setups I should check, if this information is up-to-date. If it is not used in any of those setups anymore, it is probably safe to delete. I.e. deleting it would not be problematic, because I can check that nothing directly depends on it.

I'm e.g. thinking about training jobs, where the checkpoints are maybe used by other setups.

Or about some generated dataset.

Etc.

michelwi: I am still not 100% convinced, but I can follow your reasoning

michelwi (approved):

PR #314, opened 2026-09-13
Fix task_state engine selection, avoid duplicate LocalEngine tasks

albertz:
EngineSelector.task_state selected the engine from the raw task.rqmt(), while submit() selects it from get_rqmt(), which applies the submit history and gs.check_engine_limits (that may set rqmt["engine"]). A task routed to another engine by check_engine_limits was therefore looked up in the default engine, reported as unknown, and resubmitted every manager cycle while it waited in the LocalEngine queue.
Every resubmission queued another copy, and once the LocalEngine had free slots, the same task id started several times at once
("Could not delete ... from waiting queue. This should not happen!").

task_state now resolves the engine via get_rqmt(task, task_id, update=False). LocalEngine.submit_call ignores a task id that is already queued or running, so a wrong state query can no longer start duplicates.


sisyphus/engine.py
```diff
@@ -283,7 +283,13 @@ def get_job_used_resources(self, current_process):
 
     def task_state(self, task, task_id):
         """Return state of task"""
-        return self.get_used_engine_by_rqmt(task.rqmt()).task_state(task, task_id)
+        # Select the engine the way submit() does:
+        # get_rqmt applies the submit history and gs.check_engine_limits, which may set rqmt["engine"].
+        # With the raw task.rqmt(), a task routed by check_engine_limits was looked up in the default engine,
+        # reported unknown, and resubmitted every cycle,
+        # so the same task id ran several times at once as soon as the local engine had free slots.
```
michelwi: Not sure if we need a comment here at all, but if you want some, then lets make it shorter and more concise; I don't think we need the history of the bug that was fixed here.

```suggestion
        # Respect changed requirements from submit history and gs.check_engine_limits
```

michelwi (approved):

